PR #2, opened 2018-01-11
WIP: Will create an updated _sysconfigdata.py module as part of install.

joakim-hove:
This will update the `_sysconfigdata.py` module with actual values for `prefix` and `exec-prefix`. Will fire of a test on 'bleeding' to see if it works at all.

I know it aint pretty ;-(

### DIFF
--- a/bin/kmd
+++ b/bin/kmd
@@ -10,6 +10,22 @@ import logging
 
 import komodo
 
+
+def filter_site(site_file, prefix):
+    with open(site_file) as f:
+        lines = f.readlines()
+
+    os.unlink(site_file)
+
+    with open(site_file, 'w') as f:
+        for line in lines:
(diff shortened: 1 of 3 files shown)


joakim-hove:
> Will fire of a test on 'bleeding' to see if it works at all.

Ehhh - that was not so simple; will let this dangle here for now - but hope we can merge something along this direction.

joakim-hove:
OK - have built this as *bleeding* now - superficially seems to work.

pgdr:
Will you let me know if you need anything from me on this issue?

joakim-hove:
> Will you let me know if you need anything from me on this issue?

Thank - I am doing laps on Jenkins; all the time thinking "Now it will work". If you have opinions on the general approach - which has admittedly been a bit in flux - I am happy to hear,

PR #3, opened 2018-01-20
Added protocols fs-cp and fs-ln.

joakim-hove:
The protocol used to create a 'linked' distribution is called `nfs`  and works like:
```sh
cp --recursive --synbolic-link src target
```
have added `fs-ln` as an alias for that protocol name, and also added the new protocol `fs-cp` which actually copies.


pgdr:
Merged in 19f6d5f

PR #4, opened 2018-01-20
Will not set PYTHONPATH in enable script

joakim-hove:


pgdr:
Hmm, I'm just wondering now what happens if I would like to run Komodo without building Python.

Then I would actually _like_ to have my pythonpath extended ...  ideas?

joakim-hove:
>Hmm, I'm just wondering now what happens if I would like to run Komodo without building Python. 
Then I would actually like to have my pythonpath extended ... ideas?

Ahhh  - my head spins - ; as I see it it is not really possible to achieve this in a clean way if you want to use komodo to distribute packages which *other Python distribution* naturally distribute without the use of `PYTHONPATH`

PR #8, opened 2018-02-22
Export komodo release

joakim-hove:
Just a suggetsion - will support the komodo release as an environment variable.

### DIFF
--- a/enable.m4
+++ b/enable.m4
@@ -1,2 +1,3 @@
 define(`komodo_prefix', `prefix')
 define(`komodo_pyver', `pyver')
+define(`komodo_release', `release')
--- a/bin/kmd
+++ b/bin/kmd
@@ -50,6 +50,7 @@ def main(args):
             f.write(komodo.shell(['m4 enable.m4',
                            '-D komodo_prefix={}'.format(tmp_prefix),
                            '-D komodo_pyver={}'.format('2.7'),
+                           '-D komodo_release={}'.format(args.release),
                            tmpl]))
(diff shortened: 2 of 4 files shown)


pgdr:
For which purpose, and will you export `stable` or (e.g.) `2018.01`?

Should users expect `/base/path/${KOMODO_RELEASE}/root/bin/exec` to work similarly as `exec`?

joakim-hove:
This is just a suggestion - not important. I have just added `wlog` calls to the local script - to get a feeling of how succesfull the process of pushing people over to komodo is - curently the logging call is just:
```bash
wlog komodo start
```
but it would have been nice to have:
```bash
wlog komodo $KOMODO_RELEASE
```

The current approach will be based on the build argumet, so that will be e.g. `2018.02` and not `testing`.

> Should users expect /base/path/${KOMODO_RELEASE}/root/bin/exec to work similarly as exec?

I have mnot thought of that - but I guess so?!

enable.m4
```diff
@@ -1,2 +1,3 @@
 define(`komodo_prefix', `prefix')
 define(`komodo_pyver', `pyver')
+define(`komodo_release',`release')
```
pgdr: missing comma, but other than that, feel free to merge

joakim-hove:
@pgdr : can you merge this - or give me the rights to do so?

pgdr:
Sorry!  I've added the entire team as collaborators now … thought that was default.

PR #18, opened 2018-09-17
Use symlink in prefix directory

joakim-hove:


bin/kmd
```diff
@@ -68,20 +68,13 @@ def main(args):
...
-    print('Installing {} to {}'.format(args.release, args.prefix))
-    install_root = os.path.join(args.prefix, args.release, 'root')
-
-    komodo.shell('{1} {0} .{0} {0}'.format(args.release, args.renamer))
-    komodo.shell('rsync -a .{} {}'.format(args.release, args.prefix), sudo = args.sudo)
-
-    if os.path.exists('{1}/{0}'.format(args.release, args.prefix)):
-        komodo.shell('{2} {0} {0}.delete {1}/{0}'.format(args.release, args.prefix, args.renamer),
-               sudo = args.sudo)
+    tmp_release = str(uuid.uuid4())
```
andreabrambilla: wouldn't it be a bit more informative to use something like `args.release` followed by date and time?

joakim-hove: Yes - it is now changed to use `arg-$timestamp` - not very readable; but well suited for sorting.

komodo/link.py
```diff
@@ -0,0 +1,56 @@
...
+            if ts:
+                if not release in release_list:
+                    release_list[release] = []
+
+                release_list[release].append((ts, entry))
+
+    for release,dir_list in release_list.items():
+        if os.path.exists(release):
+            if not os.path.islink(release):
+                raise Exception("Sorry - {} already exists - and is not a symblic link. This must be fixed manually".format(release))
```
pgdr: These two `if`s should be `and`ed in one if and the Exception should possibly be IO (and no need to say sorry here)

bin/kmd
```diff
@@ -68,20 +67,16 @@ def main(args):
     if args.dry_run:
         return
 
-    print('Installing {} to {}'.format(args.release, args.prefix))
-    install_root = os.path.join(args.prefix, args.release, 'root')
-
-    komodo.shell('{1} {0} .{0} {0}'.format(args.release, args.renamer))
```
pgdr: if `renamer` is no longer in use (perhaps it is, I don't know) then it should be deleted.

komodo/link.py
```diff
@@ -0,0 +1,56 @@
...
+def update_links(prefix):
+    # The first step is to scan the prefix directory and inspect all the
+    # directories in the folder. The directory names are split, and a
+    # dictionary:
+    #
+    #     release_list = {"2018.03" : [(1234, "2018.03-1234"),
+    #                                  (5678, "2018-03-5678")],
+    #                     "2018.10" : [(561, "2018.10.561"),
+    #                                  (777, "2018.10-777")]}
+    #
```
pgdr: I would prefer the timestamp is something that would be lexicographically sortable (i.e. not a running number, but a more complex date string, or at the very least 0-padded)

komodo/link.py
```diff
@@ -0,0 +1,56 @@
...
+    # This is dictionary where the release names like '2018.03' and '2018.10'
+    # are used as keys, the values in the dictionary are lists of tuples
+    # (timestamp, path_name).
+    ts_now = time.mktime( time.localtime() )
+    release_list = {}
+    for entry in os.listdir(prefix):
+        if os.path.islink(entry):
+            continue
+
+        if os.path.isdir(entry):
```
pgdr: please flip `if` and add `continue` to be consistent with the line above.  Also, I think `for→if→if→if` is too deep.

komodo/link.py
joakim-hove: Fixed

bin/kmd
joakim-hove: It is not in use - I have removed it.

komodo/link.py
joakim-hove: Did not understand 100% what you mean here - but I have rewritten the second `if` to have an early `continue`

komodo/link.py
joakim-hove: It liked the unix timestamp approach - but I have zero padded it to make it sortable in the shell.

PR #19, opened 2018-09-17
Print cwd in shell prompt

joakim-hove:
A small debugging aid.


joakim-hove:
This is trivial - and a big aid when debugging Jenkins output. Merging.

PR #20, opened 2018-09-17
Pass ddest prefix

joakim-hove:
Hack to be able to depend on opm-common

### DIFF
--- a/komodo/build.py
+++ b/komodo/build.py
@@ -39,6 +39,15 @@ def rpm(pkg, ver, path, prefix, *args, **kwargs):
         shell('rsync -a usr/* .')
         shell('rm -rf usr')
 
+
+# When running cmake we pass the option -DDEST_PREFIX=fakeroot, this is an
+# absolute hack to be able to build opm-common and sunbeam with the ~fakeroot
+# implementation used by komodo.
+#
+# See sunbeam/CMakeLists.txt for a more detailed description of the issue.
+# When/if the opm project updates the generated opm-common-config.cmake to work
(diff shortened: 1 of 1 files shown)


komodo/build.py
```diff
@@ -56,9 +56,10 @@ def cmake(pkg, ver, path, prefix, builddir,
              '-DBUILD_SHARED_LIBS=ON',
              '-DCMAKE_PREFIX_PATH={}'.format(fakeprefix),
              '-DCMAKE_MODULE_PATH={}/share/cmake/Modules'.format(fakeprefix),
-             '-DCMAKE_INSTALL_PREFIX={}'.format(prefix)
+             '-DCMAKE_INSTALL_PREFIX={}'.format(prefix),
+             '-DDEST_PREFIX={}'.format(fakeroot)
```
pgdr: Can you add a trailing `,` and a comment `  #  hack, see also Statoil/sunbeam#89`  ?

The `DDEST_PREFIX` will create a warning in all non-sunbeam cmake projects, right?

joakim-hove: OK - I have added more comments describing the issue, both here and in particular here: https://github.com/Statoil/sunbeam/pull/89.

Have not added a trailing `,` - I really don't like that technique.

pgdr: Fair enough.  The trailing comma is just so that the changes (diff) only touch the actually changed lines.  But nevermind.  Approved!

joakim-hove: Good - thank you; I agree the technique is "diff friendly" - but when actually reading the code I think it looks ugly. Anyway - in the mind of the beholder.

PR #24, opened 2018-11-06
[WIP] Add func for checking new Python versions

pgdr:
don't know if we want to merge this ever


markusdregi:
And the motivation would be that we could then get the latest version of a package from `pip` to see whether or not we are up to date? I would even argue that it would be nice to make it possible to point on a release and have it list all `pip` packages that are not up to date, with the version number in the release and the newest from `pip` displayed.

I think this is a good idea! Should we in the spirit of good software craftsmanship have tests for this stuff?

epa095:
Sounds vaguely like pip-tools

Issue #16, opened 2018-07-16
The version of cwrap reported when used from komodo is wrong (default):


joakim-hove:
```
CMake Warning at python/cmake/Modules/find_python_package.cmake:68 (message):
  Python package cwrap too old.  Wanted 1, found Loading cwrap

  0.0.0
```

PR #31, opened 2019-06-07
Remove depth from git fetch

lars-petter-hauge:
Version for an application is wrong when scm_version is used and the
depth arguments is included. The last applicable tag will not be cloned
and as such the version given would be 0.0.dev+..

When looking through the issues the following popped up: https://github.com/equinor/komodo/issues/16 I believe this issue can be closed as this is the reason this PR exists (although correct behaviour was observed prior to the `depth` keyword being included, and that is much later than the issue date)


markusdregi:
LGMT :+1:

PR #33, opened 2019-08-05
Revert inplace

pgdr:
We have settled on using a symlink based approach rather than the inplace deploy, this means that we have some dead code, which should be removed (it will still live in git history).

This reverts three commits that added the inplace deploy functionality.


markusdregi:
Looks good to me 👍 Will you remove the option from the Jenkins job also then, while you are at it?

Issue #32, opened 2019-08-02
Add ($KOMODO_RELEASE) to the csh prompt


pgdr:
When you source Komodo in bash today you get the release as a prefix for your `PS1`, it would be nice to have the same in csh.

For bash, it is essentially

```bash
export PS1="(${KOMODO_RELEASE}) ${PS1}"
```

See the files [enable.in](https://github.com/equinor/komodo/blob/master/enable.in) (bash) and [enable.csh.in](https://github.com/equinor/komodo/blob/master/enable.csh.in) (csh).

PR #46, opened 2019-11-07
Add check for KOMODO_RELEASE in csh

oyvindeide:
In bash enable there is a check making sure the users don't source different versions of `komodo` in the same session, this was not present in `csh`, added the equivalent here.


oyvindeide:
Resolves #32

markusdregi:
Let us wait until January. Rumors have it that someone is implementing deactive, and this would be much nicer to introduce at the same time ;)

oysteoh:
Will be resolved by #49

PR #50, opened 2019-12-12
Add link job that can cross link

jondequinor:
Will replace the current link job, and extend it so that you can cross link (e.g. from prog to project).

### DIFF
--- a/jenkins/link.sh
+++ b/jenkins/link.sh
@@ -1,0 +1,42 @@
+#!/bin/bash
+
+set -e
+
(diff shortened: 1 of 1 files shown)


markusdregi:
Looks good to me 👍 I don't think we will ever need the cross linking will we? As I see it the `/project/res` links are one-time creations that will just live as is on top of what is on `/prog`. So, we can either say that we only operate on `/prog` and start pretending that `/project` does not exist, or we can keep it like this ;)

PR #54, opened 2019-12-17
Added auto-style-formatter + validator

### LINKED ISSUE #52: "Prettier" option
Common CI tools for Python are [`black`](https://github.com/psf/black) for code style formatting, and [`pylint`](https://github.com/PyCQA/pylint) for static code analysis. JavaScript has a similar quite standard pair: [`prettier`](https://github.com/prettier/prettier) for code style formatting, and [`eslint`](https://github.com/eslint/eslint) for static code analysis.

`black` / `prettier` automatically change the code style to comply with some predefined style. Static code analysis tools however reports error, and typically never try to fix them automatically (as that would in many cases involve some assumptions that could change the program behavior).

`komodo` has ["static analysis"](https://github.com/equinor/komodo/blob/master/komodo/lint.py) which can report missing version definitions, undefined dependency, undefined maintainer etc.

Would also be nice with an automatic "style formatter" ensuring that e.g. the lists are indeed alphabetically sorted, empty lines are consistent, no space before `:` in `yaml` dictionaries, no `''` around strings, consistently use two space indentation etc. This style formatter could, like `black` and `prettier`, have a `--check` option which could be used in CI to ensure that also a preferred style is followed.

anders-kiaer:
Closes #52.

### DIFF
--- a/komodo/prettier.py
+++ b/komodo/prettier.py
@@ -1,0 +1,157 @@
+from __future__ import print_function
+
+import os
+import re
+import sys
+import argparse
+import functools
+
+# On Python3, StringIO can come from standard library io:
+from ruamel.yaml.compat import StringIO
+import ruamel.yaml
+
+
+def repository_specific_formatting(empty_line_top_level, yaml_string):
+    '''Transform function to ruamel.yaml's dump function. Makes sure there are
+    only empty lines inbetween different top level keys (if empty_line_top_level
+    is True, otherwise no empty lines).
+    '''
+
+    yaml_string = re.sub(r'\n+', r'\n', yaml_string)  # Remove all empty lines
+
+    if empty_line_top_level:
+        yaml_string = re.sub(  # Add one empty line between each package
+            r'(\n[^\#][^\n]*)\n([^\s])', r'\1\n\n\2', yaml_string
+        )
+
+    return yaml_string
+
+
+def is_repository(config):
+    '''Returns `False` if the configuration corresponds to a Komodo release
+    (all config elements below top level key are strings). Returns `True` if
+    it corresponds to a _repository_ (all config elements below top level key are
+    themselves dictionaries).
+
+    Raises ValueError if inconsistent throughout the config.
+    '''
+
+    # For Python2+3 compatibility. On Python3-only, use isinstance(x, str)
+    # instead of isinstance(x, basestring).
+    try:
+        basestring
+    except NameError:
+        basestring = str # No basestring on Python3
+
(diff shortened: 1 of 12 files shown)


komodo/prettier.py
```diff
@@ -0,0 +1,149 @@
...
+    '''Takes in a string corresponding to a YAML Komodo configuration, and returns
+    the corresponding prettified YAML string.'''
+
+    ruamel_instance = ruamel.yaml.YAML()
+    ruamel_instance.indent(  # Komodo prefers two space indendation
+        mapping=2, sequence=4, offset=2
+    )
+    ruamel_instance.width = 1000  # Avoid ruamel wrapping long
+
+    config = ruamel_instance.load(yaml_input_string)
```
oysteoh: We should try to be concrete on which kind of errors we expect and by reraising a systemexit with the error message the output is much more precise and informative for the user.

```
try:
    config = ruamel_instance.load(yaml_input_string)
except ruamel.yaml.constructor.DuplicateKeyError as e:
    raise SystemExit(str(e))
```

komodo/prettier.py
```diff
@@ -0,0 +1,149 @@
...
+
+    parser = argparse.ArgumentParser(
+        description=(
+            'Check and/or format the Komodo configuration files. '
+            'Takes in any number of yml files, which could be e.g. the main '
+            'Komodo repository and an arbitrary number of releases. '
+            'Throws a hard error if the same package is defined multiple times.'
+        )
+    )
+    parser.add_argument(
```
oysteoh: to avoid hard exit and weired stacktraces when inputting a folder, you could do something like this
```
def is_valid_file(parser, arg):
    if not os.path.isfile(arg):
        parser.error("{} is not a file!".format(arg))
    else:
        return arg

parser.add_argument(
    "files",
    type=lambda path: is_valid_file(parser, path),
    nargs="+",
    help="One or more files to format/check",
)
```

oysteoh:
Looks good, and seems to work more or less as expected. There is some ugly stacktraces on different errors, some of them should probably be handled in a more informative way. Take a look at my comments and consider if you agree or not!

komodo/prettier.py
anders-kiaer: Agree - easier (and shorter) user feedback on duplicate keys increases user friendliness. Added suggestion 👍

anders-kiaer:
Thanks @oysteoh 👍 Good suggestions on removing stack traces for common user input errors, and instead provide simplified messages. More or less copied in your suggestions - see if you agree 🙂

PR #71, opened 2020-01-28
Use build pip

markusdregi:


komodo/build.py
```diff
@@ -101,9 +101,10 @@ def rsync(pkg, ver, pkgpath, prefix, *args, **kwargs):
     shell(['rsync -am', kwargs.get('makeopts'), '{}/'.format(pkgpath), kwargs['fakeroot'] + prefix])
 
 
-def pip(pkg, ver, pkgpath, prefix, dlprefix, *args, **kwargs):
+def pip_install(pkg, ver, pkgpath, prefix, dlprefix, *args, pip='pip', **kwargs):
```
sondreso: `pip='pip'` needs to come before `args` and `kwargs` in python 2

komodo/build.py
```diff
@@ -101,9 +101,10 @@ def rsync(pkg, ver, pkgpath, prefix, *args, **kwargs):
...
     # pip should be in the tmp root, which is already added to PATH
```
sondreso: This comments does not make sense after this commit

PR #74, opened 2020-02-04
Remove args when pip installing git downloaded libs

oysteoh:
Not sure why these are present - they was used when pip installing dependencies on our packages - or at least they were present in the code. Will try to remove these and make the installation more like how normal pip packages is installed. 

Don't think this will fix the dirty version issue - but at least it's a good thing removing fluff we don't need (hopefully)

### DIFF
--- a/setup-py.sh
+++ b/setup-py.sh
@@ -36,11 +36,7 @@ while test $# -gt 0; do
     shift
 done
 
-PYTHONEXECUTABLE=$PREFIX/bin/python
-
-pip install . \
-    --global-option build --global-option=--executable=$PYTHONEXECUTABLE \
-    --root $FAKEROOT                                                     \
-    --no-deps                                                            \
+pip install .           \
+    --root $FAKEROOT    \
+    --no-deps           \
     --prefix $PREFIX
-


setup-py.sh
```diff
@@ -36,11 +36,7 @@ while test $# -gt 0; do
     shift
 done
 
-PYTHONEXECUTABLE=$PREFIX/bin/python
-
-pip install . \
-    --global-option build --global-option=--executable=$PYTHONEXECUTABLE \
-    --root $FAKEROOT                                                     \
-    --no-deps                                                            \
+pip install .           \
```
markusdregi: Is it not so that it builds towards the Python version in the release instead of the one in the current environment?

sondreso: But then why are we not passing these arguments to pip install for packages fetched from pypi?

oysteoh: my question exactly... wanted to make this pip installation more like the other one... i don't know

markusdregi: Good question. Perhaps we should? For now we are building with a rather similar Python as we are targeting. So perhaps this is the reason it has worked out so far?

markusdregi: Either way, I definitively support unification!

oysteoh:
This seems to work along with https://github.com/equinor/komodo-releases/pull/204 .

PR #87, opened 2020-02-19
Make it possible to extract dependency graph

oysteoh:


oysteoh:
@jondequinor is this the start on what you hoped for with your issue?

jondequinor:
It most definitely is. Only nitpick currently is that I would like `--out` to default to STDOUT. So if no `--out` just `print()` the output

PR #90, opened 2020-02-27
Add linting of package status files

sondreso:
Needed for https://github.com/equinor/scout/issues/136

### DIFF
--- a/komodo/lint_package_status.py
+++ b/komodo/lint_package_status.py
@@ -1,0 +1,90 @@
+#!/usr/bin/env python
+
+import argparse
+import os
+import sys
+
+import yaml
+
+
+VALID_VISIBILITY = ["public", "private"]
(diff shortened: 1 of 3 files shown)


komodo/lint_package_status.py
```diff
@@ -0,0 +1,94 @@
...
+    if repository_set.difference(package_status_set):
+        raise SystemExit(
+            "The following packages are specified in the repository file, but not in the package status file: {}".format(
+                list(repository_set.difference(package_status_set))
+            )
+        )
+
+    errors = []
+    for package, status in package_status.items():
+        if not is_valid_option(status, "visibility", VALID_VISIBILITY):
```
ManInFez: could be simplified to 
```suggestion
      if not status.get('visibility') in VALID_VISIBILITY
```

Issue #98, opened 2020-04-29
When updating stable, deprecated should also be updated


dotfloat:
`deprecated` points to the previous `stable` release. We should update both when symlink mode is `stable`.

markusdregi:
This was not sufficient earlier as the compute cluster got the new stable immediately, but I support implementing this now as the cluster is currently getting the same environment as the client 👍

dotfloat:
Decided not to update automatically. Rather, we should update deprecated a week prior to updating stable.

PR #99, opened 2020-04-29
Update deprecated symlink when stable is updated

dotfloat:
Resolves: #98 

I'm not particularly confident that the added tests cover everything though.

### DIFF
--- a/komodo/symlink/suggester/configuration.py
+++ b/komodo/symlink/suggester/configuration.py
@@ -49,6 +49,11 @@ class Configuration(object):
                 else:
                     self.links[link] = repr(release)
         elif mode == "stable":
+            # Add/update deprecated link iff stable exists and has changed
+            if self.links.get(link) != release.month_alias():
+                dlink = link.replace("stable", "deprecated")
+                self.links[dlink] = self.links[link]
+
             self.links[release.month_alias()] = repr(release)
             self.links[link] = release.month_alias()
         else:
--- a/tests/test_suggester.py
+++ b/tests/test_suggester.py
@@ -161,6 +161,7 @@ def test_get_concrete_release(conf, link, concrete):
     "links": {
         "2015.12-py27": "2015.12.03-py27",
         "2016.01-py27": "2016.01.00-py27",
+        "deprecated-py27": "2015.12-py27",
         "stable-py27": "2016.01-py27"
     }
 }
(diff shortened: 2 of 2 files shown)


komodo/symlink/suggester/configuration.py
```diff
@@ -49,6 +49,11 @@ def update(self, release, mode):
                 else:
                     self.links[link] = repr(release)
         elif mode == "stable":
+            # Add/update deprecated link iff stable exists and has changed
+            if self.links.get(link) != release.month_alias():
```
xjules, line 53: Why not just `self.links[link]` ?

komodo/symlink/suggester/configuration.py
```diff
@@ -49,6 +49,11 @@ def update(self, release, mode):
...
+                dlink = link.replace("stable", "deprecated")
+                self.links[dlink] = self.links[link]
+
             self.links[release.month_alias()] = repr(release)
```
xjules, line 57: Well, this question doesn't belong to your PR, but why it's called `month_alias` when it returns essential `Y.M-PYVER` ?

tests/test_suggester.py
```diff
@@ -276,6 +300,29 @@ def test_suggest_symlink_configuration():
     )
 
 
+def test_deprecated_noop_suggestion():
+    config = """{"links": {
+"2050.02-py58": "2050.02.00-py58",
```
xjules, line 305: You really think in 2050 we will still call it python? :)

tests/test_suggester.py
```diff
@@ -276,6 +300,29 @@ def test_suggest_symlink_configuration():
...
+        release="2050.02.00-py58",
+        mode="stable",
+        symlink_conf_path="foo.json",
+        joburl="http://job",
+        jobname="job",
+    )
+
+    repo.create_pull.assert_not_called()
+
+    assert suggest_symlink_configuration(args, repo) is None
```
xjules, line 323: Not really sure if I understand this test, ie. shouldn't we test actually the content of the suggestion too?

komodo/symlink/suggester/configuration.py
dotfloat: Because `self.links[link]` will throw an exception if no `stable` exists. The alternative is
`link in self.links[link] and self.links[link] != release.month_alias()`, but my version is terser.

komodo/symlink/suggester/configuration.py
dotfloat: Because `month_alias` is the alias to the actual release and represents the month. @jondequinor can probably answer better.

jondequinor: Yeah, this is correct. No release should be without the python version postfix.

dotfloat:
We have decided to update `deprecated` a week prior to updating `stable`. Closing.

Issue #97, opened 2020-04-22
Komodo incompatible with GNOME on RHEL7


oyvindeide:
Komodo is incompatible with GNOME on RHEL7 if sourced in a  `.*rc`-file, because that is sourced from `.*_profile`, and the python version overwrites the system default.

Known issues:
 - [ ] gnome-tweaks
 - [ ] frozen login screen

berland:
By using
```
tty -s && source /prog/res/komodo/testing-py36/enable
```
the problem seemingly goes away, at least I can login, komodo is correctly activated  in terminals. Not tested through ERT/LSF (where this sourcing then probably will not happen, but you said this is fixed by other means?)

dotfloat:
Found out that `LD_LIBRARY_PATH` is the problem. Other projects (https://github.com/pyenv/pyenv, maybe conda) can insert their own Pythons without issue. The problem is that libraries conflict with the system. Initialising komodo with the following fixes both GNOME log in and KDE font sizes:
```bash
_ld=$LD_LIBRARY_PATH
source /prog/res/komodo/bleeding-py36/enable
export LD_LIBRARY_PATH=$_ld
```
While `python` is Komodo's.

I propose the following solution:
1. Komodo's `enable` only sets `PATH` and `PS1`.
2. `PATH` points to `root/shims` instead of `root/bin`
3. All files in `root/shims` are wrappers for files in `root/bin`. The difference is that they set `LD_LIBRARY_PATH` (if not set before)

Alternatively, we can fix the RPATH for all libraries that use Komodo's libs so that we don't need to set `LD_LIBRARY_PATH`.

dotfloat:
Here's a list of libraries (only `.so` files) in komodo that have missing dependencies:
https://gist.github.com/dotfloat/88f50045398a170d0642efd1a31fbee8

Notice that all of these are either our software or are built by us. Some oddities: PyQt5's Python extension libraries all have their RPATH set to `/project/res/komodo/repository/qt5/root/lib` and these are the ones that get loaded. That is, the libraries `/prog/res/komodo/testing-py36/root/lib/libQt5*.so` are entirely unused.

We should compile things with RPATH enable when possible: this would have to be a change in the `kmd` tool. For software that we rsync in, we can use https://github.com/NixOS/patchelf to patch the RPATH.

markusdregi:
This has been solved

PR #104, opened 2020-05-11
Patch RPATH in Komodo's ELF binaries

dotfloat:
Resolves: #97 

Use https://github.com/Equinor/patchelf instead of https://github.com/NixOS/patchelf . The upstream clears old RPATHs, which doesn't play well when `ld` tries to compress the entries. In Stavanger, you can do `export PATCHELF_EXEC=/project/oompf/patchelf/patchelf`

### DIFF
--- a/tests/test_elf.py
+++ b/tests/test_elf.py
@@ -46,7 +46,9 @@ def test_dyn(tmpdir):
 
 def test_symlink(tmpdir):
     """Symlinks may point to ELF files outside of the komodo release, so it is
-imperative that we do not touch them"""
+    imperative that we do not touch them
+
+    """
     with tmpdir.as_cwd():
         os.symlink("/bin/bash", "symlink")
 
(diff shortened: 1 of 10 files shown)


tests/test_elf.py
```diff
@@ -0,0 +1,144 @@
...
+def test_dyn(tmpdir):
+    """Compile a shared object and test it"""
+    with tmpdir.as_cwd():
+        _compile("gcc -shared {code}")
+
+        assert os.path.isfile("a.out")
+        assert elf.is_valid_elf_file("a.out")
+
+
+def test_symlink(tmpdir):
```
sondreso, line 47: Just to be certain, should we make a test that symlinks a folder, and assert that the files inside that folder are not counted as valid files?

dotfloat: Yep! See enclosed commit.

sondreso: Alright, if the tests are green I have nothing against building a test release :+1:

PR #107, opened 2020-05-19
Compare release and tag maturity

hnformentin:
#107 
test_lint_maturity.py under construction

### DIFF
--- a/komodo/lint_maturity.py
+++ b/komodo/lint_maturity.py
@@ -1,0 +1,257 @@
+#!/usr/bin/env python
+
+import argparse
+import os
+import sys
+import yaml
+from packaging.version import Version, parse, InvalidVersion
+import warnings
+
+
+_INVALID_TAGS = {
+    "a": ["invalid"],
+    "b": ["a", "invalid"],
+    "rc": ["a", "b", "invalid"],
+    "stable": ["a", "b", "rc", "invalid"],
+    "exception": ["a", "b", "rc", "invalid"],
+}
+
+
+def print_system_exit_message(system_exit_msg):
+
+    if system_exit_msg != "":
+        raise SystemExit(system_exit_msg)
+
+
+def print_warning_message(system_warning_msg):
+
+    if system_warning_msg != "":
+        warnings.warn(system_warning_msg, UserWarning)
+
+
+def msg_packages_invalid(
+    release_basename, release_version, count_tag_invalid, dict_tag_maturity
+):
+
+    exit_msg = ""
+    exit_msg += (
+        release_basename
+        + " has "
+        + str(count_tag_invalid)
+        + " packages with invalid maturity tag.\n"
+    )
+
+    for tag in _INVALID_TAGS[release_version]:
+
+        if len(dict_tag_maturity[tag]) > 0:
+            exit_msg += (
+                "\tTag " + tag + " packages: " + str(dict_tag_maturity[tag]) + "\n"
+            )
+
+    return exit_msg
+
+
+def msg_packages_exception(release_basename, dict_tag_maturity):
+
+    msg_exception = ""
+
+    if len(dict_tag_maturity["exception"]) > 0:
+        msg_exception += release_basename + ", exception list of packages:\n"
+        msg_exception += "\t" + str(dict_tag_maturity["exception"]) + " \n"
+
+    return msg_exception
+
+
+def count_invalid_tags(dict_tag_maturity, invalid_tags):
+
+    count_tag_maturity = {tag: len(dict_tag_maturity[tag]) for tag in dict_tag_maturity}
+    invalid_tags_count = sum(count_tag_maturity[tag] for tag in invalid_tags)
+
+    return invalid_tags_count
+
+
+def get_release_type(version_string):
+
+    try:
+        version = Version(version_string)
+        release_type = "stable" if version.pre is None else version.pre[0]
+    except InvalidVersion:
+        release_type = "invalid"
+
+    return release_type
+
+
+def get_packages_info(packages_dict, tag_exceptions_package):
+
+    dict_tag_maturity = {
+        "a": [],
+        "b": [],
+        "rc": [],
+        "stable": [],
+        "exception": [],
+        "invalid": [],
+    }
+
+    for package_name, package_version in packages_dict.items():
+
+        if package_name not in tag_exceptions_package:
+            release_version_package = get_release_type(package_version)
+        else:
+            release_version_package = "exception"
+        dict_tag_maturity[release_version_package].append(
+            (package_name, package_version)
+        )
+
+    return dict_tag_maturity
+
+
+def read_yaml_file(file_path):
+
+    with open(file_path) as yml_file:
+        loaded_yaml_file = yaml.safe_load(yml_file)
+
+    return loaded_yaml_file
+
+
+def msg_release_exception(release_basename, release_version):
+
+    msg_exception = ""
+
+    if release_version == "exception":
+        msg_exception += (
+            release_basename + " not lint because it is in the exception list.\n"
+        )
+
+    return msg_exception
+
+
+def get_release_version(release_basename, tag_exceptions_release):
+
+    release_cleanname = release_basename.split("-")[0]
+
+    if release_cleanname not in tag_exceptions_release:
+        release_version = get_release_type(release_cleanname)
+    else:
+        release_version = "exception"
+
+    return release_version
+
+
+def run(files_to_lint, tag_exceptions):
+
+    system_exit_msg = ""
+    system_warning_msg = ""
+
+    for file_to_lint in files_to_lint:
+
+        release_basename = os.path.basename(file_to_lint)
+        release_version = get_release_version(
+            release_basename, tag_exceptions["release"]
+        )
+        system_warning_msg += msg_release_exception(release_basename, release_version)
+
+        if release_version == "invalid":
+            system_exit_msg += (
+                release_basename + " is incompatible with version name.\n"
+            )
+        else:
+            packages_dict = read_yaml_file(file_path=file_to_lint)
+            dict_tag_maturity = get_packages_info(
+                packages_dict, tag_exceptions["package"]
+            )
+            count_tag_invalid = count_invalid_tags(
+                dict_tag_maturity, _INVALID_TAGS[release_version]
+            )
+
+            system_warning_msg += msg_packages_exception(
+                release_basename, dict_tag_maturity
+            )
+
+            if count_tag_invalid > 0:
+                packages_msg = msg_packages_invalid(
+                    release_basename,
+                    release_version,
+                    count_tag_invalid,
+                    dict_tag_maturity,
+                )
+
+                if release_version != "exception":
+                    system_exit_msg += packages_msg
+                else:
+                    system_warning_msg += packages_msg
+
+    print_warning_message(system_warning_msg)
+    print_system_exit_message(system_exit_msg)
+
+
+def get_files_to_lint(release_folder, release_file):
+
+    if release_folder is None:
+        files_to_lint = [release_file]
+    else:
+        files_to_lint = [
+            os.path.join(release_folder, file_in_dir)
+            for file_in_dir in os.listdir(release_folder)
+        ]
+
+    return files_to_lint
+
+
+def define_tag_exceptions(tag_exception_arg):
+
+    if os.path.isfile(tag_exception_arg[0]):
+        tag_exceptions = read_yaml_file(file_path=tag_exception_arg[0])
+    elif tag_exception_arg[0] == "":
+        tag_exceptions = {"release": [], "package": []}
+    else:
+        exit_msg = str(tag_exception_arg) + " is not a valid file."
+        raise SystemExit(exit_msg)
+
+    return tag_exceptions
+
+
+def get_parser():
+
+    parser = argparse.ArgumentParser(description=("Lint the maturity of packages."))
+    parser.add_argument(
+        "--tag_exceptions",
+        nargs="+",
+        default=[""],
+        help="One file with exceptions for (a) release tag and (b) packages tag.",
+    )
+
+    group = parser.add_mutually_exclusive_group(required=True)
+    group.add_argument(
+        "--release_file",
+        type=lambda arg: arg
+        if os.path.isfile(arg)
+        else parser.error("{} is not a valid file".format(arg)),
+        help="Release file",
+    )
+    group.add_argument(
+        "--release_folder",
+        type=lambda arg: arg
+        if os.path.isdir(arg)
+        else parser.error("{} is not a valid directory".format(arg)),
+        help="File with all package tags named as release version.",
+    )
+
+    return parser
+
+
+def main():
+
+    parser = get_parser()
+    args = parser.parse_args()
+
+    tag_exceptions = define_tag_exceptions(tag_exception_arg=args.tag_exceptions)
+
+    files_to_lint = get_files_to_lint(
+        release_folder=args.release_folder, release_file=args.release_file
+    )
+
+    run(files_to_lint, tag_exceptions)
+
+
+if __name__ == "__main__":
+    main()
--- a/tests/test_lint_maturity.py
+++ b/tests/test_lint_maturity.py
@@ -1,0 +1,444 @@
+import pytest
+import os
+import sys
+from komodo import lint_maturity
+import warnings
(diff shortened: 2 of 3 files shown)


komodo/lint_maturity.py
```diff
@@ -0,0 +1,104 @@
...
+
+def define_release_version(release_file):
+    try:
+        Version(release_file)
+        release_version = (
+            "None"
+            if parse(release_file).pre is None
+            else str(parse(release_file).pre[0])
+        )
+    except:
```
oyvindeide, line 46: Blanket `except` is usually not a good idea, you want to catch specific errors, otherwise it can lead to some quite strange behavior.

tests/test_lint_maturity.py
```diff
@@ -0,0 +1,89 @@
...
+         Tag b packages: {'package_b1': 'v3.1.b1', 'package_b2': '3.1.b1+ert'}
+         Tag Invalid packages: {'package_IV1': 'testing/2020.3/rc1', 'package_IV2': '4.19.19-py3', 'package_IV3': '5.13.1-src', 'package_IV4': 'v1.2.3-nolibeclreq'}
+bleeding-py27.yml is incompatible with version naming 
+"""
+
+FOLDER_TEST = "/test_lint_maturity/"
+
+
+def create_test_files(RELEASE_SAMPLE, FILE_NAMES_SAMPLE, FOLDER_TEST):
+    folder_name = os.getcwd() + FOLDER_TEST
```
oyvindeide, line 50: I would consider running this in a `tmp`  folder. `pytest` has a built-in `tmpdir` fixture you can use.  I would try to avoid using try/except in tests, because although this one immediately raises an error, you don`t really know why it raises said error.

komodo/lint_maturity.py
```diff
@@ -0,0 +1,104 @@
...
+                count_tag_maturity[key] for key in INVALID_TAGS[release_version]
+            )
+            
+            system_exit=system_exit+release_file_basename+" has "+str(count_tag_invalid)+" packages with invalid maturity tag. \n"
+            if count_tag_invalid > 0:
+                for key in INVALID_TAGS[release_version]:
+                    if len(dict_tag_maturity[key]) > 0:
+                        system_exit=system_exit+"         Tag "+ key+ " packages: "+ str(dict_tag_maturity[key])+'\n'
+
+    raise SystemExit(system_exit)
```
oyvindeide, line 85: This function does a lot, and I would consider splitting it up into several smaller functions, specifically regarding the file reading. A side effect is that it might make it easier to test as well, and easier to understand what is going on.

tests/test_lint_maturity.py
```diff
@@ -0,0 +1,89 @@
...
+def test_lint_maturity_folder():
+
+    list_files = create_test_files(RELEASE_SAMPLE, FILE_NAMES_SAMPLE, FOLDER_TEST)
+
+    with pytest.raises(SystemExit) as exit_info:
+        lint_maturity.run(list_files)
+
+    for file_name in list_files:
+        os.remove(file_name)
+    os.rmdir(os.path.dirname(list_files[0]))
```
oyvindeide, line 73: Is this for clean up? By running this in a `tmp` folder you wont have to worry about that, and it is one less place where your test can unexpectedly fail.

komodo/lint_maturity.py
hnformentin: In the end this exception was not required, but I learned that I need to import the error from the package. Many thanks.

tests/test_lint_maturity.py
hnformentin: That was a very good suggestion. Thanks for it.

komodo/lint_maturity.py
hnformentin: I tried to review and it was indeed possible to split and make more readable. Thanks for highlighting this.

tests/test_lint_maturity.py
```diff
@@ -0,0 +1,118 @@
...
+            RELEASE_SAMPLE.rsplit("\n", 4)[0], [FILE_NAMES_SAMPLE[1]]
+        )
+    try:
+        with pytest.raises(SystemExit) as exit_info, warnings.catch_warnings(
+            record=True
+        ) as warning_info:
+            lint_maturity.run(file_name, TAG_EXCEPTIONS)
+        pytest.fail("SystemExit was raised when it should not.")
+    except:  # check
+        pass
```
oyvindeide, line 105: This is a bit confusing to me, because it looks like you want it to raise an exception, when in fact you´re doing the excact opposite, you could do something like this:
```suggestion
    with pytest.warns(UserWarning) as record:
        lint_maturity.run(file_name, TAG_EXCEPTIONS)

    assert record[0].message.args[0] == "2020.02.a1-py27.yml, exception list of packages:\n"
```

Though looking at it, you might not want it to raise that warning at all in this case?

oyvindeide: If you don´t want it to raise any warnings you could confirm with:
```suggestion
def test_lint_maturity_pass(tmpdir):

    with tmpdir.as_cwd():
        file_name = create_test_files(
            RELEASE_SAMPLE.rsplit("\n", 4)[0], [FILE_NAMES_SAMPLE[1]]
        )
    with pytest.warns(None) as record:
        lint_maturity.run(file_name, TAG_EXCEPTIONS)

    assert not record
```

hnformentin: Thanks Øyvind! I understand this may be confusing. 
Here I would like to test that nothing is raised (warning and/or SystemExit) when all packages have appropriate maturity for a giving release. 
It seems that the suggested code would not test this, but perhaps renaming would be a way to improve and avoid confusion. 
Would you have a better suggestion for the name test_lint_maturity_pass to represent this objective?

oyvindeide: That´s what I thought after thinking some, my second suggestion will assert that, but will currently fail as a warning was raised when I ran your code.

oyvindeide: And it should be enough to assert that no warnings are raised, as a `SystemExit` will cause the test to fail.

komodo/lint_maturity.py
```diff
@@ -0,0 +1,162 @@
...
+def get_release_type(version_string):
+    try:
+        version = Version(version_string)
+        release_type = "stable" if version.pre is None else version.pre[0]
+    except InvalidVersion:
+        release_type = "Invalid"
+    return release_type
+
+
+def get_release_info(file_to_lint, tag_exceptions_release):
```
oyvindeide, line 56: Great that you split your function up, now it is much easier to read. Another thing that would be great now that you´ve done that would be to write unit tests for these smaller functions. This function could for example be tested using `pytest parameterized`:
```
@pytest.mark.parametrize(
    "input_file,exceptions,clean_release,version,warning",
    [("/a/random/path/2020.02.a1-py27", [], "2020.02.a1-py27", "a", "")],
)
def test_release_info(input_file, exceptions, clean_release, version, warning):
    release_name, package_version, warning_msg = lint_maturity.get_release_info(
        input_file, exceptions
    )
    assert release_name == clean_release
    assert package_version == version
    assert warning_msg == warning
```
Then you can add different test cases to that.

komodo/lint_maturity.py
```diff
@@ -0,0 +1,206 @@
...
+
+import argparse
+import os
+import sys
+import yaml
+from packaging.version import Version, parse, InvalidVersion
+import warnings
+
+
+INVALID_TAGS = {
```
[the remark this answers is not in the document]
hnformentin: Very good, I should probably get familiar with PEP 8 in general.
Will also apply for _create_tmp_test_files in the test script. Thanks!

komodo/lint_maturity.py
```diff
@@ -0,0 +1,206 @@
...
+    "a": ["invalid"],
+    "b": ["a", "invalid"],
+    "rc": ["a", "b", "invalid"],
+    "stable": ["a", "b", "rc", "invalid"],
+    "exception": ["a", "b", "rc", "invalid"],
+}
+
+
+def get_parser():
+    parser = argparse.ArgumentParser(description=("Lint the maturity of packages."))
```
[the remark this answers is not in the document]
hnformentin: Good point! I was thinking a different logic but it makes sense your suggestion so I implemented it. Thanks.

komodo/lint_maturity.py
```diff
@@ -0,0 +1,206 @@
...
+        release_type = "invalid"
+    return release_type
+
+
+def get_release_version(release_basename, tag_exceptions_release):
+    release_cleanname = release_basename.split("-")[0]
+    release_version = (
+        get_release_type(release_cleanname)
+        if release_cleanname not in tag_exceptions_release
+        else "exception"
```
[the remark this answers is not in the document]
hnformentin: Agree and modified :-) Thanks!

komodo/lint_maturity.py
```diff
@@ -0,0 +1,206 @@
...
+                    release_version,
+                    count_tag_invalid,
+                    dict_tag_maturity,
+                )
+
+    print_warning_message(system_warning_msg)
+    print_system_exit_message(system_exit_msg)
+
+
+def main():
```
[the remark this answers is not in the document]
hnformentin: You have spotted a new function to be defined, thanks! :-) 
Now we are smaller functions in main() and unit tests for all of them.
An integration test is in run(), so I believe we do not need to add a integration test to main().
Would you agree with that?

hnformentin:
> This is looking pretty good, I have some tests to go through, but some comments in the mean time. I also ran it on a release file, and this was the result:
> `Tag invalid packages: ['pyqt5', '5.13.1-src']` is that expected? It is valid `semver`.

Yes, '-' makes it an invalid semver. For my understanding, we use both pyqt5 and pyqt5-sip with some kind of adaptation, so these packages were included in the exception list ;-)

tests/test_lint_maturity.py
```diff
@@ -0,0 +1,300 @@
...
+\tTag a packages: [('package_a1', 'v3.1.a1')]
+\tTag b packages: [('package_b1', 'v3.1.b1')]
+\tTag rc packages: [('package_rc1', 'v3.1.rc1')]
+\tTag invalid packages: [('package_IV1', '5.13.1-src')]
+"""
+
+
+def _create_tmp_test_files(release_sample, file_names_sample):
+
+    folder_name = os.getcwd() + "/test_lint_maturity/"
```
oyvindeide, line 127: Best to use `os.path.join()` when creating paths, that way it is platform agnostic.

tests/test_lint_maturity.py
```diff
@@ -0,0 +1,300 @@
...
+    count_tag_invalid = EXPECTED_N_INVALID_TAGS[release_version]
+
+    exit_msg = lint_maturity.msg_packages_invalid(
+        release_basename,
+        release_version,
+        count_tag_invalid,
+        EXPECTED_DICT_TAG_MATURITY,
+    )
+
+    assert exit_msg in EXPECTED_SYSTEMEXIT
```
oyvindeide, line 155: This is a general comment that goes for all your tests, I think they would be much easier to read if you change them to be more explicit, like this:
```suggestion
@pytest.mark.parametrize(
    "release_basename, release_version, count_tag_invalid",
    [
        ('2020.02.01-py27.yml', 'stable', 4),
        ('2020.02.a1-py27.yml', 'a', 1),
        ('2020.02.b1-py27.yml', 'b', 2),
        ('2020.02.rc1-py27.yml', 'rc', 3)
    ]
)
def test_msg_packages_invalid(release_basename, release_version, count_tag_invalid):

    exit_msg = lint_maturity.msg_packages_invalid(
        release_basename,
        release_version,
        count_tag_invalid,
        EXPECTED_DICT_TAG_MATURITY,
    )

    assert exit_msg in EXPECTED_SYSTEMEXIT
``
I would even prefer including `EXPECTED_DICT_TAG_MATURITY` and `EXPECTED_SYSTEMEXIT`, but that might be harder. It makes it much easier to read what you are actually testing.

File: komodo/lint_maturity.py (hunk 10 above)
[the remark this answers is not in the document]
Comment (oyvindeide) [reply komodo/lint_maturity.py]: Not quite, although most of your logic is contained in the `run` function, and your code is well tested, it would be good to have an integration test of the full workflow as well, so all your code is actually tested. Theoretically I could now delete everything in `main` and the tests would still be green.

File: komodo/lint_maturity.py [hunk 13]
```diff
@@ -0,0 +1,255 @@
...
+        help="File with all package tags named as release version.",
+    )
+
+    return parser
+
+
+def main():
+
+    parser = get_parser()
+    args = vars(parser.parse_args())
```
oyvindeide, line 245: Why do you want to run it through `vars` by the way? I would think it would be better not to do that.

komodo/lint_maturity.py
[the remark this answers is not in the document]
oyvindeide: If you would also like to add a unit test to `main`, it could look something like this:
```
def test_main(monkeypatch):
    parser_mock = mock.Mock()
    args = mock.Mock()
    args.tag_exceptions = "an_exception"
    parser_mock.parse_args.return_value = args
    get_parser_mock = mock.Mock(return_value=parser_mock)
    monkeypatch.setattr(lint_maturity, "get_parser", get_parser_mock)

    tag_exceptions_mock = mock.Mock(return_value=["list", "of", "exceptions"])
    monkeypatch.setattr(lint_maturity, "define_tag_exceptions", tag_exceptions_mock)

    files_to_lint_mock = mock.Mock(return_value=["list", "of", "files"])
    monkeypatch.setattr(lint_maturity, "get_files_to_lint", files_to_lint_mock)

    run_mock = mock.Mock()
    monkeypatch.setattr(lint_maturity, "run", run_mock)

    assert tag_exceptions_mock.called_once_with("an_exception")
    assert files_to_lint_mock.called_once_with(args)
    assert run_mock.called_once_with(["list", "of", "files"], ["list", "of", "exceptions"])

    lint_maturity.main()
```

komodo/lint_maturity.py
hnformentin: Thanks! 
It was a (perhaps non-elegant) way to answer a previous review comment, to test the function ´main()´.
I tried to break the function further by defining get_files_to_lint, and avoided testing large strings from ´main´ outcomes (which turned to be difficulty because of /t and /n).
To test get_files_to_lint, it appeared that having the variable arg as a dictionary was more feasible compared to creating a parsed object.
Why is it not good? Would you have a suggestion to create and test a function that call a parser object?
Many thanks!

oyvindeide: Sure, the best way would be to use a `mock`, for example like this (also see my unit test of `main` example):

```
args_mock = mock.Mock()
args_mock.release_folder = "some/folder"
args_mock.release_file = "some_fname"
```

Note that you could bypass this whole problem by not passing in `args` to `get_files_to_lint`, which might actually be preferable, as you want a function to work on as basic types as possible, and not "know" too much.

tests/test_lint_maturity.py
hnformentin: Thanks! 
Even if the same structure/data is used multiple times in the code?
I thought it would be too much wordy to be explicit for some values, e.g. RELEASE_FILE_NAMES is used 7 times, EXPECTED_RELEASE_VERSION is used 8, TAG_EXCEPTIONS 16 times.
Would it be ok if I only let explicit what is used only once?

oyvindeide: In terms of `EXPECTED_DICT_TAG_MATURITY` and `EXPECTED_SYSTEMEXIT` it makes more sense to have global ones. Is it possible to make the tests even more generic though, so they dont expect the same? That might not make sense though so feel free to disregard.

komodo/lint_maturity.py
oyvindeide: My preference would be just that, to pass two strings.

tests/test_lint_maturity.py
hnformentin: I was a little skeptical when I read but I now think your suggestion was very good. many thanks!
After the review, what is tested is closer to the tests and I also add the name of the arguments explicitly in some functions to make it easier to read.
It is perhaps a little wordy, but that is because we have a test per type of release (stable, a,b,rc, ...) most of the time.

komodo/lint_maturity.py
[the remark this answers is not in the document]
hnformentin: This _mock_ concept is quite new for me. Many thanks. 
I tried to implement this test. Please let me know if you have concerns on that.

PR #115, opened 2020-06-08
Have pip force-reinstall packages always

dotfloat:
Pip, if it finds an existing package, doesn't install it again even if we've specified `--root`/`--prefix`. This was discovered when we updated the version of `six` to be the same as latest. Komodo has `six` as one of its dependencies and so pip discovered that it was already installed in the temporary Komodo virtualenv, and so didn't install it into the release.

I still need to have Jenkins use Python 3 for running Komodo regardless of what the bundle python will be, but that isn't critical yet.


dotfloat:
All tests (incl hm tutorial) passed for bleeding-py27 and -py36. Merging.

PR #119, opened 2020-06-20
Delay setting LD_LIBRARY_PATH until executing something from bin/

dotfloat:
Currently we are setting `LD_LIBRARY_PATH` when komodo is enabled. If it is sourced in the user's `.bashrc`, it breaks a lot of system-managed software, most notably GNOME (Our `libxkbcommon.so` that we ship with komodo causes a segfault in `gnome-shell`).

This is yet another proposal on how to fix this. Instead of changing `LD_LIBRARY_PATH` when komodo is enabled, we delay it until the user actually runs some program that we provide. Thus, `LD_LIBRARY_PATH` affects only the program the user wanted and any child processes that it may spawn.

We do this thusly:
1. Build a release as normal
2. At the end, rename `bin/` to `libexec/`
3. For each entry in `libexec/`:
    1. Create a 0755 entry in `bin/` with the same name, containing the following bash script:
```bash
#!/bin/bash -eu
root=$(dirname $0)/..
prog=$(basename $0)
if [ -z "${_KOMODO_SHIM:-}" ]; then
  if [ -z "${LD_LIBRARY_PATH:-}" ]; then
    export LD_LIBRARY_PATH=$root/lib:$root/lib64
  else
    export LD_LIBRARY_PATH=$root/lib:$root/lib64:$LD_LIBRARY_PATH
  fi
  export _KOMODO_SHIM=$prog
fi
$root/libexec/$prog "$@"
```

---

While this solves the problem of GNOME segfaulting, it isn't without its issues:
1. All of our integration tests need to be rewritten. This is something that likely must be done with the other attempts too.
    1. They all use `virtualenv --system-site-packages` to extend the release with pytest and other testing packages
    2. Some (*libecl*, *libres*) use `ctest`, requiring setting the `LD_LIBRARY_PATH` manually

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -60,13 +60,16 @@ script:
                 --release travis                                            \
                 --renamer rename.ul                                         \
                 --pip $PIP
-  - tree /tmp/pfx
+  # Build libkmd.so for testing
+  - ci/travis/build_lib.sh /tmp/pfx/travis
 
+  - tree /tmp/pfx
   # Use built komodo
   - source /tmp/pfx/travis/enable
   - which python
   - python --version
   - python -c "import numpy;print(numpy.__file__)"
+  - ci/travis/test_import_lib.py
 
   # Test that everything makes sense
   - '[[ "$(which python)" == "/tmp/pfx/travis/root/bin/python" ]]'
--- a/ci/travis/build_lib.sh
+++ b/ci/travis/build_lib.sh
@@ -1,0 +1,10 @@
+#!/bin/bash -eu
+
+prefix=$1
+
+# Compile
+echo 'char*foo(){return"bar";}' > /tmp/lib.c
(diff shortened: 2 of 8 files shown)


sondreso:
> While this solves the problem of GNOME segfaulting, it isn't without its issues:
...
Some (libecl, libres) use ctest, requiring setting the LD_LIBRARY_PATH manually

I don't see this as that big of a problem. We control theses tests, so could we not set the LD_LIBRARY_PATH in the script that starts the tests instead?

dotfloat:
Yes, after a lot of discussions the solution here is to have Komodo be Python-only and any non-Python libraries will not be supported and the workflow may be broken as we see fit. So our tests can be fixed, but if anyone else is using their C programs against komodo, then it's their problem.

PR #123, opened 2020-06-26
Handle virtualenv built python

jondequinor:


komodo/build.py
```diff
@@ -77,58 +79,72 @@ def cmake(pkg, ver, path, prefix, builddir,
...
            '--root {}'.format(kwargs['fakeroot']),
            '--prefix {}'.format(prefix),
-           '--force-reinstall',
            '--no-index',
            '--no-deps',
+           '--ignore-installed',
            '--find-links {}'.format(dlprefix),
-           kwargs.get('makeopts', '')]
+    ]
+    cmd.append(kwargs.get('makeopts', ''))
```
dotfloat, line 120: Why use append instead of just having this as the last list elem as in the previous version?

komodo/build.py
```diff
@@ -77,58 +79,72 @@ def cmake(pkg, ver, path, prefix, builddir,
...
 
-def make(pkgfile, repofile, prefix = None,
-                            dlprefix = None,
-                            builddir = None,
-                            jobs = 1,
-                            cmk = 'cmake',
-                            pip = 'pip',
-                            fakeroot = '.'):
+
+def make(pkgfile, repofile, data,
```
dotfloat, line 139: I think we should create a single `Configuration` object that has most of this information (function args) rather than passing a million different ones separately.

We don't have to do it in this PR, but I'd like to join in writing it. Wasn't entirely sure how to do this pythonically so I stopped trying to implement it myself.

tests/test_cli.py
```diff
@@ -0,0 +1,66 @@
...
+        "--workspace",
+        str(tmpdir),
+    ]
+
+    # Komodo expects the renamer from util-linux. On Debian/Ubuntu, this
+    # renamer has been renamed to rename.ul.
+    if os.path.exists("/usr/bin/rename.ul"):
+        sys.argv.extend(["--renamer", "/usr/bin/rename.ul"])
+
+    sys.argv.extend([arg for arg in args])
```
dotfloat, line 54: `sys.argv.extend(args)`?

In general I think modifying `sys.argv` at runtime is not a sign of quality software. Then again, this is komodo, so nevermind.

komodo/__init__.py
```diff
@@ -35,14 +35,14 @@ def fixup_python_shebangs(prefix, release):
...
-        try:
+    try:
+        for bin_ in os.listdir(binpath):
             with open(os.path.join(binpath, bin_), 'r') as f:
                 shebang = f.readline().strip()
             if _is_shebang(shebang):
                 bins_.append(bin_)
-        except Exception as err:
-            print('Exception in reading bin: %s' % err)
+    except Exception as err:
```
dotfloat, line 44: `catch(Exception e) { e.printStackTrace(); }`

An exception here should fail the build imo.

komodo/build.py
jondequinor: Yeah, there was something in between there that has since been removed. Good catch

komodo/build.py
jondequinor: Yeah, let's do that pairwise. I'm not sure either.

tests/test_cli.py
jondequinor: You're right, it's bad. I'll change this so that I use `_main` instead, and then do argparse testing elsewhere referencing the parser explicitly.

jenkins/matrix-build/Jenkinsfile
```diff
@@ -0,0 +1,64 @@
...
+                            build job: 'x-build-some-other', parameters: [
+                                string(name: 'RH_VERSION', value: RH_VERSION),
+                                string(name: 'PYTHON_VERSION', value: PY_VERSION),
+                                string(name: 'MATRIX_FILE_BASE', value: params.MATRIX_FILE_BASE),
+                                booleanParam(name: 'overwrite', value: params.overwrite)
+                            ], wait: true
+                        }
+                        post {
+                            success {
+                                // build job: 'komodo-suggest-symlink', parameters: [
```
ManInFez, line 45: I suppose this is to commented back in?

jondequinor: Yeah, it's time.

komodo/cli.py
```diff
@@ -2,152 +2,210 @@
...
-                                         builddir = args.tmp,
-                                         jobs     = args.jobs,
-                                         cmk      = args.cmake,
-                                         pip      = args.pip,
-                                         fakeroot = fakeroot)
-        komodo.shell('mv {} {}'.format(args.release + tmp_prefix, args.release))
-        komodo.shell('rmdir -p --ignore-fail-on-non-empty {}'.format(
-            os.path.dirname(args.release + tmp_prefix)))
+        komodo.make(
+            args.pkgs,
```
ManInFez, line 35: why are we not using argument names on the first three?

komodo/cli.py
```diff
@@ -2,158 +2,206 @@
...
     if args.build and not args.install:
         sys.exit(0)
 
     # create the enable script
-    for tmpl,target in [('enable.in','enable'), ('enable.csh.in', 'enable.csh')]:
-        if not os.path.exists(tmpl):
-            logging.warn('Could not find template %s, skipping.' % str(tmpl))
-            continue
+    for tmpl, target in [("enable.in", "enable"), ("enable.csh.in", "enable.csh")]:
+        # if not os.path.exists(tmpl):
```
ManInFez, line 59: either comment back in or remove

komodo/cli.py
jondequinor: Because that's the way it was designed. Maybe add it to the pile? https://github.com/equinor/komodo/issues/125

komodo/cli.py
jondequinor: Right, we don't want to warn here, we want to fail hard. Will fix.

komodo/data/__init__.py
```diff
@@ -0,0 +1,17 @@
+import os
+
+
+class Data(object):
+    def __init__(self, extra_data_dirs=None):
+        """Create a Data object responsible for giving you the filez you need.
+        Takes a list of extra_data_dirs to search."""
+        self._dirs = [os.path.dirname(__file__)]
+        if extra_data_dirs is not None:
+            self._dirs.extend(extra_data_dirs)
```
ManInFez, line 10: should we validate that the extra dirs are existing?

komodo/data/build__python-virtualenv.sh
```diff
@@ -0,0 +1,30 @@
...
+            export VIRTUALENV_INTEPRETER=$1
+            ;;
+        *)
+            export OPTS="$OPTS $1"
+            ;;
+    esac
+    shift
+done
+
+ $VIRTUALENV -p $VIRTUALENV_INTEPRETER -v --always-copy --never-download --no-setuptools --no-pip --no-wheel ${FAKEROOT}${PREFIX} 1>&2
```
ManInFez, line 30: ${FAKEROOT}${PREFIX} should result in a path? what if missing /?

jondequinor: Will fix.

komodo/data/__init__.py
jondequinor: done

komodo/local.py
```diff
@@ -0,0 +1,11 @@
+import operator
+from jinja2 import Template
+
+
+def write_local_activators(data, defs, local, local_csh):
```
ManInFez, line 5: We need to to use this as well, right? could not find any usage except in test..

jondequinor: The call is `local.write_local_activators` and that is used around line 70 in `cli.py`

tests/test_release_transpiler.py
```diff
@@ -27,6 +27,7 @@ def test_build_release_matrix(tmpdir):
         assert os.path.isfile(new_release_file)
         with open(new_release_file) as f:
             release_matrix = yaml.safe_load(f)
+            print(release_matrix)
```
ManInFez, line 30: print needed?

ManInFez (approved):
All in all looks good! merger is held off until 2020.09 stable is in place

vars/komodo.groovy
```diff
@@ -1,255 +1,175 @@
...
+            booleanParam name: 'overwrite', defaultValue: false,
+                description: 'Whether or not to overwrite if build already exist'
+
+            string name: 'CODE_GIT_FORK', defaultValue: 'equinor',
+                description: 'The fork to the get the komodo build system from'
+            string name: 'CODE_GIT_REF', defaultValue: 'allow_no_python_package',
+                description: ' The branch to get the komodo build system from'
+            string name: 'CONFIG_GIT_FORK', defaultValue: 'equinor',
+                description: 'The fork to get the the komodo releases from'
+            string name: 'CONFIG_GIT_REF', defaultValue: 'rhel7-py2',
```
dotfloat, line 84: May wanna change the defaults before merging this PR.

vars/komodo.groovy
```diff
@@ -1,255 +1,175 @@
...
+                                repository.yml                                         \
+                                --download                                             \
+                                --build                                                \
+                                ${params.deploy ? '--install' : '--dry-run'}           \
+                                --jobs 6                                               \
+                                --release ${getReleaseName()}                          \
+                                --tmp tmp                                              \
+                                --cache cache                                          \
+                                --prefix ${params.PREFIX}                              \
+                                --pip ${komodoPath}/boot/build-env/bin/pip             \
```
dotfloat, line 159: Unnecessary. The correct `pip` is already in `PATH` via `runkmd.sh`. Same should be true for `virtualenv`.

komodo/data/jenkins.sh
```diff
@@ -0,0 +1,46 @@
+#!/bin/bash
```
dotfloat, line 1: What is the purpose of this file?

vars/komodo.groovy
jondequinor: Unfortunately, some packages will require that `PATH` be set to the builds's `bin` path. For builtin python, that `pip` is going to be old and quite possibly not work at all.

komodo/data/jenkins.sh
jondequinor: You tell me, it came from `komodo-releases`. https://github.com/equinor/komodo-releases/blob/master/jenkins.sh

vars/komodo.groovy
jondequinor: Actually, I think it's necessary. I will demonstrate.

PR #135, opened 2020-07-22
shim: Use the 'exec' bash builtin

dotfloat:
Previously, bash started the shimmed program as a separate process. This change makes it so that the shimmed program takes over the bash process **and** the program thinks it's located in `bin/` (the `-a $0` part). If you run python, it'll:
1. Replace `bash` process with `../libexec/python`
2. Python's `sys.executable` will think it's `../bin/python`. Therefore, any process that uses `sys.executable` will correctly use the shimmed version that sets up the variables instead of directly.

`exec -a $0`

### DIFF
--- a/komodo/shim.py
+++ b/komodo/shim.py
@@ -4,10 +4,8 @@ from textwrap import dedent
 
 shim_template = dedent("""\
     #!/bin/bash
-    root=$(dirname $0)/..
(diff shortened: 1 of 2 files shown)


dotfloat:
Shims are dead :sob: Closing PR

PR #140, opened 2020-08-07
Enable message of the day

ManInFez:
resolves equinor/ert#735
depends on https://github.com/equinor/komodo-releases/pull/620


dotfloat:
This is really good. I think it's better if the script goes over every release in `/prog/res/komodo` and matches them.

If this script goes over every release, then we'll be able to do things like apply messages to `stable`, which will just be put in whatever `stable` points to.

A rule like the following should apply to `2020.08.01-py27` if that is pointed to by `stable-py27`.
```
'stable*':
  messages:
    - thank-you-for-using-stable.txt
```


For a rule like
```
'*-py27':
  messages:
    - py27-is-deprecated.txt
```
could match both `stable-py27` and `2020.01.00-py27` that `stable-py27` points to. In that case only 1 such message should be added.

dotfloat:
Also, do we want to have messages inline in the YAML file? It'll cut down on the number of files in git for trivial messages.

komodo/post_messages.py
```diff
@@ -0,0 +1,154 @@
...
+
+    parser.add_argument(
+        "--motd-db", "-m", required=True, help="YML file defining the messages.",
+    )
+
+    parser.add_argument(
+        "--komodo-prefix",
+        "-k",
+        required=True,
+        help="Path to folder holding komodo-releases.",
```
dotfloat, line 76: Ambiguous: Path to `https://github.com/equinor/komodo-releases` or `/prog/res/komodo`? Perhaps `"Path to komodo releases root (eg. /prog/res/komodo)"`. Also make it the default.

dotfloat (approved):
LGTM

komodo/post_messages.py
```diff
@@ -0,0 +1,154 @@
...
+    scripts = []
+    messages = []
+    inline = []
+    for key in motd_db:
+        if fnmatch.fnmatch(release_name, key):
+            scripts = (
+                scripts + motd_db[key].get("scripts")
+                if motd_db[key].get("scripts") is not None
+                else scripts
+            )
```
sondreso, line 20: `scripts.extend(motd_db.get(key,[]))`?

sondreso, line 20: `### scripts.extend(motd_db[key].get("scripts", []))`?

komodo/post_messages.py
```diff
@@ -0,0 +1,154 @@
...
+
+    for msg in messages:
+        filename = hashlib.md5(msg.encode()).hexdigest()
+        filename = "0Z" + filename  # for orderings sake
+        with open(os.path.join(dst_path, filename), "w") as new_file:
+            new_file.write(msg)
+
+
+def copy_files(file_list, dst_path, src_path):
+    if os.path.isdir(dst_path):
```
ManInFez, line 49: should not happen, as they are cleared in the beginning

komodo/post_messages.py
```diff
@@ -0,0 +1,154 @@
...
+                inline + motd_db[key].get("inline")
+                if motd_db[key].get("inline") is not None
+                else inline
+            )
+
+    return scripts, messages, inline
+
+
+def create_inline_messages(messages, clear_old_messages, dst_path):
+    if clear_old_messages:
```
ManInFez, line 36: should not happen, as they are cleared in the beginning

sondreso (approved):
Except for the comments this LGTM as well :rocket:

PR #141, opened 2020-08-12
add reverse dependency graph extractor

mortalisk:
Can print a textual hierachical ascii thingy, or a dot graph description. Can also call "dot" and "display" to visualise the graph.
Example:
`python komodo/reverse_dep_graph.py releases/bleeding-py36.yml repository.yml --display_dot`

### DIFF
--- a/komodo/reverse_dep_graph.py
+++ b/komodo/reverse_dep_graph.py
@@ -1,0 +1,148 @@
+#!/usr/bin/env python3
+
+import argparse
+import os
+import sys
+import yaml
+import subprocess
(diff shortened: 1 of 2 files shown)


oysteoh:
new code should be blacked - at least i would appreciate it :+1:

oysteoh:
would also as much as blacking appreciate a test verifying the expected behavior :+1:  other than that - i think this tool is an awesome additon to komodo :medal_sports:

oysteoh (approved):
:rocket:

Issue #95, opened 2020-03-12
Validate release names


markusdregi:
The current pipeline requires each release file to be postfixed with python version, i.e. `-py27` or `-py36`. However, this is not validated before merging. The auto tooling then just silently ignores the newly added release.

PR #151, opened 2020-09-04
Lint release file names

dotfloat:
Resolves: #95 

Add linting that will check if a release name ends with one of: `-py27`, `-py36`, `-py27-rhel6`, `-py27-rhel7`, `-py36-rhel6`, `-py36-rhel7`.

### DIFF
--- a/komodo/lint.py
+++ b/komodo/lint.py
@@ -1,16 +1,18 @@
 #!/usr/bin/env python
 from __future__ import print_function
+
+import os
 import argparse
-from pkg_resources import parse_version
 import logging
 import yaml as yml
+from pkg_resources import parse_version
 
 from collections import namedtuple
 
 kerr = namedtuple('KomodoError',
                   ['pkg', 'version', 'maintainer', 'depends', 'err'])
 
-report = namedtuple('LintReport', ['maintainers', 'dependencies', 'versions'])
+report = namedtuple('LintReport', ['release_name', 'maintainers', 'dependencies', 'versions'])
 
 MISSING_PACKAGE = 'missing package'
 MISSING_VERSION = 'missing version'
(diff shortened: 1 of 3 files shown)


tests/test_lint.py
```diff
@@ -28,12 +28,37 @@
...
+
+    for invalid in (
+        "bleeding",
+        "bleeding.yml",
+        "2020.01.01",
+        "2020.01.00.yml",
+        "/home/anyuser/komodo-releases/releases/2020.01.00.yml",
+        "bleeding-py27",
+        "bleeding-rhel6.yml"
+    ):
```
sondreso, line 57: This seems like a perfect case for [pytest.mark.parameterize ](https://docs.pytest.org/en/stable/parametrize.html)!

komodo/lint.py
```diff
@@ -43,6 +45,20 @@ def _validate(pkg, ver, repo):
     return _kerr(pkg=pkg, version=ver, maintainer=repo[pkg][ver]['maintainer'])
 
 
+def lint_release_name(pkgfile):
+    relname = os.path.basename(pkgfile)
+    found = False
+    for py_suffix in "-py27", "-py36":
+        for rh_suffix in "", "-rhel6", "-rhel7":
+            if relname.endswith(py_suffix + rh_suffix + ".yml"):
```
sondreso, line 53: Should we not allow `relname.endswith(py_suffix + ".yml")`, or are we planning to switch everything over before merging this?

dotfloat: `rh_suffix` can be, among other things, `""`.

tests/test_lint.py
dotfloat: Debateable. I kind of dislike `pytest.mark.parametrize` and this codebase uses it very unsuccessfully. I did change my test to use it though, so it's up to you.

komodo/lint.py
sondreso: You are totally correct, I misread :disappointed:

sondreso (approved):
LGTM!

tests/test_lint.py
sondreso: Yes, might not make much difference in this case, but think we can keep it as is now!

PR #164, opened 2020-11-06
Unset DESTDIR when pip install .

dotfloat:
DESTDIR is used by Komodo and other build packagers to compile the software when there is no access `/usr`. When installing Python packages, this is entirely superfluous, and actually breaks skbuild's build process, which gets incorrectly prefixed with the value of DESTDIR.


ManInFez (approved):
:bacon:

PR #168, opened 2020-12-03
Add package_name field to pip_install

eivindjahren:
Implement the repository field "pypi_package_name" to specify name of a package in pypi when installing via `pip_install`.

### DIFF
--- a/komodo/build.py
+++ b/komodo/build.py
@@ -7,9 +7,8 @@ import os
 import sys
 from distutils.dir_util import mkpath
 
-import komodo
-
-from .shell import pushd, shell
+from komodo.package_version import strip_version
+from komodo.shell import pushd, shell
 
 flatten = itr.chain.from_iterable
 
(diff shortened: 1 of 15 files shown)


komodo/build.py
```diff
@@ -115,17 +115,31 @@ def rsync(pkg, ver, pkgpath, data, prefix, *args, **kwargs):
...
+        "--prefix {}".format(prefix),
+        "--no-index",
+        "--no-deps",
+        "--ignore-installed",
+        "--cache-dir {}".format(dlprefix),
+        "--find-links {}".format(dlprefix),
+        kwargs.get("makeopts", ""),
     ]
 
     print('Installing {} ({}) from pip'.format(pkg, ver))
```
jondequinor, line 145:
```suggestion
    print('Installing {} ({}) from pip'.format(package_name, ver))
```

komodo/build.py
```diff
@@ -115,17 +115,31 @@ def rsync(pkg, ver, pkgpath, data, prefix, *args, **kwargs):
...
-           kwargs.get('makeopts', ''),
+def pip_install(
+    pkg,
+    ver,
+    pkgpath,
+    data,
+    prefix,
+    dlprefix,
+    pip="pip",
+    package_name=None,
```
jondequinor, line 126: For `pip`, `pkg == package_name`, so drop this? Use `pkg` from this point on.

komodo/build.py
```diff
@@ -162,58 +176,79 @@ def make(pkgfile, repofile, data,
...
+        "noop": noop,
+    }
 
     for pkg, path in zip(pkgorder, pkgpaths):
         ver = pkgs[pkg]
         current = repo[pkg][ver]
-        make = current['make']
+        make = current["make"]
         pkgpath = os.path.abspath(path)
+        package_name = pkgs.get("package_name", None)
```
jondequinor, line 228: I think we should only support `package_name` for `source: pypi`, so do an assertion here.

Basically, if `package_name`, the `source` must be `pypi`.

jondequinor (commented):
There's a second install of all pip packages at the bottom of `cli.py`. You need to fix this there as well.

komodo/build.py
eivindjahren: This has now been changed

komodo/build.py
eivindjahren: Added

jondequinor:
Could you add a unit test for this? We have some testing in place already, so should be easy to extend.

jondequinor (approved):
Good stuff, thank you!

jondequinor (approved):
🥳

PR #172, opened 2020-12-07
Migrate CI to Github Actions

sondreso:


jondequinor:
In what state is this PR, @sondreso ?

sondreso:
I was trying to get rid of the use of anaconda, but realized that was very difficult. Feel free to hijack this PR or close it and start over again if you want this to be merged within this week :smile:

dotfloat:
The entire point of using anaconda is to test that `komodo` will be able to create a release for a different Python version. Doesn't using the built-in Python defeat this purpose?

.github/workflows/test.yml
```diff
@@ -44,7 +44,8 @@ jobs:
 
     - name: Full integration test
       run: |
-        ./runkmd.sh ci/travis/${RELEASE} ci/travis/repository.yml --workspace /tmp/kmd-ws --cache /tmp/kmd-cache --prefix /tmp/pfx --release travis --locations-config $(realpath ci/travis/locations.yml) --renamer rename.ul --pip boot/bintools/pip
+        which pip
+        ./runkmd.sh ci/travis/${RELEASE} ci/travis/repository.yml --workspace /tmp/kmd-ws --cache /tmp/kmd-cache --prefix /tmp/pfx --release travis --locations-config $(realpath ci/travis/locations.yml) --renamer rename.ul --pip $(which pip)
```
dotfloat, line 48: This is very incorrect! The `pip` here comes from the Python environment `kmd` runs in. Not the target Python environment. Remove `--pip $(which pip)` since it defaults to `pip` in the code, and will find the `boot/bintools/pip` instead as it should.

jondequinor: I am still investigating, hold your horse. But default `pip` here failed, so there's something wrong with the environment.

dotfloat: Ok. I'll hold on to my horse for the time being.

jondequinor:
> The entire point of using anaconda is to test that `komodo` will be able to create a release for a different Python version. Doesn't using the built-in Python defeat this purpose?

Isn't this a part of a matrix with three separate python versions? Different than what?

jondequinor:
> > The entire point of using anaconda is to test that `komodo` will be able to create a release for a different Python version. Doesn't using the built-in Python defeat this purpose?
> 
> Isn't this a part of a matrix with three separate python versions? Different than what?

Also, I consider komodo the thing that should be tested.

dotfloat:
There was once a time where in order to make a 2.7-compatible komodo release, komodo itself also had to be 2.7-compatible. Thus the Travis test built a Python 2.7 and 3.7 komodo release using a 3.6-compatible komodo.

There's currently 3 different environments that the build process juggles:
* **komodo**'s, which can be whatever Python version we choose to write `kmd` in
* **build**, which contains build tools needed to build the komodo release, such as `pip` or `cmake`
* **target**, which is the one that gets deployed as a komodo release

In particular it's important to note that while **build** and **target**'s Python versions need to be kept in sync (wholly due to `pip`), this restriction doesn't get applied to the **komodo** environment, which only houses `kmd` and friends. `bootstrap.sh` resolves this for you so that you don't need to think about it.

That said, this PR looks fine. We should consider installing a newer system Python, however, in order to test that komodo is capable of building any version, not just 3.6.

dotfloat (approved):

PR #176, opened 2020-12-17
Add AGPLv3 license file

dotfloat:


markusdregi:
May I ask what the rationale for choosing AGPL vs for instance GPL is? :)

dotfloat:
Make it as strict as possible while keeping it FOSS.

markusdregi:
In case someone wants to provide Komodo built distributions as a service in a closed-source tool ;)

However, I don't mind using AGPL as the default in a AGPL vs GPL discussion as it does really close some of the oddities with respect to GPL in these modern times...

markusdregi (approved):

Issue #181, opened 2021-01-07
Double Komodo release in csh prompt


markusdregi:
When sourcing `komodo/stable` in a cshell I get the following prompt:

> [2020.11.00-py36] [2020.11.00-py36-rhel7] [mfane@rgsnode ~] $

while in bash I get

> (2020.11.00-py36) [mfane@rgsnode ~] $

Should we unify?

dotfloat:
This is a bug, yes. I have provided a fix but unsure if it fixes stuff. It also seems odd to me that we are setting `KOMODO_RELEASE` at all in the `generic` -> `rhel7` redirect enable script...

PR #182, opened 2021-01-07
csh: Use $_PRE_KOMODO_PS1 instead of $prompt

dotfloat:
Using $prompt here causes the komodo release version to double up in the user's
PS1. The bash version of this uses $_PRE_KOMODO_PS1 and so should the csh
version.

Resolves: #181


markusdregi:
Looks like a good approach 👍 Tests are failing though...

PR #191, opened 2021-02-18
Add vulnerability checking with Snyk

sondreso:
Add security checking for komodo releases using Snyk.

### DIFF
--- a/komodo/snyk_reporting.py
+++ b/komodo/snyk_reporting.py
@@ -1,0 +1,178 @@
+import argparse
+import html
+import os
+import sys
+from typing import Any, Dict, List
+
+if sys.version_info >= (3, 7):
+    from snyk import SnykClient
+    from snyk.managers import OrganizationManager
+    from snyk.models import Vulnerability
+else:
+    OrganizationManager = Any
+    Vulnerability = Any
+
+from komodo.yaml_file_type import ReleaseDir, ReleaseFile, YamlFile
+
+_CONSOLE_VULNERABILITY_FORMAT = """\t{id}
+\t\tPackage: {package}
+\t\tVersion: {version}
+\t\tTitle: {title}
+\t\tURL: {url}
+\t\tIdentifiers: {identifiers}
+\t\tSeverity: {severity}
+"""
+
+_GITHUB_VULNERABILITY_FORMAT = (
+    "| {id} | {package} | {version} | {title} | {url} | {identifiers} | {severity} |\n"
+)
+
+
+def main() -> None:
+    api_token = os.getenv("SNYK_API_TOKEN")
+    args = parse_args(sys.argv[1:])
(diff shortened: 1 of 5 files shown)


sondreso:
pysnyk does not exist for python 3.6 🙁

komodo/snyk_reporting.py
```diff
@@ -0,0 +1,158 @@
...
+        releases=releases, repository=args.repo, org=org
+    )
+
+    if args.format_github:
+        print(format_github(vulnerabilities=vulnerabilities))
+    else:
+        print(format_console(vulnerabilities=vulnerabilities))
+
+
+if __name__ == "__main__":
```
markusdregi, line 157: Is this necessary when it is installed as a console script?

markusdregi (commented):
The code looks really neat 👏 A test would not be bad ;)

komodo/snyk_reporting.py
sondreso: No, but it eases testing slightly since you can run it in your IDE without installing. Also did it since it seems to be convention in these files, but nothing against removing it.

markusdregi: No problem 👍 Feel free to keep it...

frode-aarstad:
Looks good. Perhaps add a test for the actual Snyk API key if that is not done elsewhere

sondreso:
Good point! Think a had a test for that at some point, but it must have been lost in some refactor 😅  Will add that again!

tests/test_snyk_reporting.py
```diff
@@ -0,0 +1,129 @@
...
+    return result_mock
+
+
+@above_py37
+def test_no_api_token():
+    with pytest.raises(
+        ValueError,
+        match=r"No api token given, please set the environment variable SNYK_API_TOKEN.",
+    ):
+        vulnerabilities = snyk_main(
```
mortalisk, line 49: vulnerabilities is unused

tests/test_snyk_reporting.py
```diff
@@ -0,0 +1,129 @@
...
+            releases={}, repository={}, api_token=None, org_id="some_org_id"
+        )
+
+
+@pytest.mark.skipif(sys.version_info >= (3, 7), reason="requires Python < 3.7")
+def test_python36():
+    with pytest.raises(RuntimeError):
+        vulnerabilities = snyk_main(
```
mortalisk, line 57: vulnerabilities is unused

frode-aarstad (approved):

mortalisk (approved):
LGTMAF

tests/test_snyk_reporting.py
sondreso: I know, but I left it as this because you could easily see the interface from the tests 🙂

mortalisk: Makes sense

PR #195, opened 2021-02-25
Use package alias to get version for release doc

jondequinor:


markusdregi (approved):
Looks good according to my wits 🤷

PR #198, opened 2021-05-12
Add script for copying upgrade-proposals to a release

mortalisk:


jondequinor (commented):
If we run this on jenkins, do we need to do the git work (adding the changes, creating the commit, creating the PR) in the job directly?

I'm comparing it to https://github.com/equinor/komodo/blob/master/komodo/symlink/suggester/cli.py where the git work is done in the script itself.

mortalisk:
If you run it on Jenkins, nothing special will happen. It will not do any git anything for anyone, no matter where it might be run.

I suggest we create small tools that can build on each other. So perhaps another script can do the git things? Perhaps we can generalize the concept of, run this script, then propose a PR of the results?

jondequinor:
I see little value of such a script outside automation. I'd much rather just have the PR. But maybe recent RMs can chime in? @frode-aarstad

frode-aarstad:
The automatic creation of a PR like Jonas describes is very good improvement. 

Its not much work creating the PR manually but being able to approve the PR as release manager saves some time.

komodo/insert_proposals.py
```diff
@@ -0,0 +1,193 @@
...
+
+    pr_msg = f""":robot: Release {target} created by script
+---
+### Description
+- New Release: `{target}`
+- Based on: `{base}`
+- When: `{datetime.now()}`
+
+### Diff
+diff {base_file} {target_file}:
```
frode-aarstad, line 132: Perhaps move this into the formatted block and add $ diff

komodo/insert_proposals.py
```diff
@@ -0,0 +1,193 @@
...
+        branch=tmp_target,
+    )
+
+    # making PR
+    base_content = repo.get_contents(base_file, ref=args.git_ref)
+    diff = diff_file_and_string(
+        b64decode(base_content.content).decode(), result, base, target
+    )
+
+    pr_msg = f""":robot: Release {target} created by script
```
frode-aarstad, line 124: Perhaps remove "created by script" redundant since :robot:

frode-aarstad (approved):

PR #210, opened 2021-07-01
Use python 3 for creating venv for suggest-symlink

mortalisk:


markusdregi (approved):
This is a good idea! 😱

mortalisk:
> This is a good idea! 😱

Python 2.7 came up with it, and gave a big fat warning already.

markusdregi:
Good advice though ;)

PR #216, opened 2021-08-16
Add script that suggest upgrades from pypi

oyvindeide:


komodo/check_up_to_date_pypi.py
```diff
@@ -0,0 +1,167 @@
...
+
+def get_release(yaml, fname):
+    with open(fname, "r") as fin:
+        result = yaml.load(fin)
+    return result
+
+
+def get_packages_pypi(package_names):
+    return [
+        (package, requests.get(f"https://pypi.python.org/pypi/{package}/json"))
```
sondreso, line 29: Just out of curiosity, how long does this take? :sweat_smile:

komodo/check_up_to_date_pypi.py
```diff
@@ -0,0 +1,167 @@
...
+
+
+def yaml_parser():
+    yaml = ruamel.yaml.YAML()
+    yaml.indent(mapping=2, sequence=4, offset=2)
+    yaml.preserve_quotes = True
+    return yaml
+
+
+def get_release(yaml, fname):
```
sondreso, line 21: maybe rename to load_from_file (or something similar)?

komodo/check_up_to_date_pypi.py
```diff
@@ -0,0 +1,167 @@
...
+    return pypi_packages
+
+
+def get_upgrade_proposals(release_file, repository_file, python_version, propose_upgrade):
+    yaml = yaml_parser()
+    releases = get_release(yaml, release_file)
+    repository = get_release(yaml, repository_file)
+
+    pypi_packages = get_pypi_packages(releases, repository)
+    pypi_response = get_packages_pypi(pypi_packages)
```
sondreso, line 83: The naming here is slightly confusing, we could call the bottom one get_pypi_info or something like that instead?

komodo/check_up_to_date_pypi.py
```diff
@@ -0,0 +1,167 @@
...
+            )
+        if pypi_latest_version != komodo_version:
+            upgrade_proposals[name] = {"previous": str(komodo_version), "suggested": str(pypi_latest_version)}
+
+    if propose_upgrade and upgrade_proposals:
+        for name, version in upgrade_proposals.items():
+            suggested_version = version["suggested"]
+            if suggested_version not in repository[name]:
+                repository[name].update(
+                    {suggested_version: copy.deepcopy(repository[name][releases[name]])}
```
sondreso, line 105: If I have understood this correct, then maybe this reads a little bit better? :thinking: 
```suggestion
                    {suggested_version: copy.deepcopy(repository[name][version["previous"]])}
```

sondreso (commented):
Very nice :clap:  Other then my small comments/question this looks really good!

komodo/check_up_to_date_pypi.py
oyvindeide: Shorter than you (or I at least) would think. I started shooting of more at the time, but it was not really needed I felt.

komodo/check_up_to_date_pypi.py
oyvindeide: I agree that would be better, but the version there has been sanitized (to remove `+py3` or what we might find appropriate to add), so will not fit directly. Will try to improve readability.

oyvindeide: And at the same time, add a test for that `+py3` thing

oyvindeide:
Did a small refactor to improve readability, and added a full integration test.

sondreso (approved):
LGTM!

komodo/check_up_to_date_pypi.py
oyvindeide: Dont get me wrong, it is quite slow, but acceptably slow (in my opinion)

PR #218, opened 2021-08-18
Add script that checks dependency conflicts

oyvindeide:


oyvindeide:
Is this worth merging?

jondequinor:
I think that it's not, since it's not considering the actual, built distribution. I think the easier and more correct approach would be to have a `pip check` integrated into the komodo build. What do you think?

jondequinor:
In either case, added https://github.com/equinor/komodo/issues/219

oyvindeide:
> I think that it's not, since it's not considering the actual, built distribution. I think the easier and more correct approach would be to have a pip check integrated into the komodo build process. What do you think?

I see your point, and having it at the end of the build process would make sense, but they are not exclusive. I see the use case of this more when bumping packages from pypi and needing to check dependencies. It will shorten the feedback loop significantly as opposed to having it as a part of the build process. It is of course not a fail safe, as some internal packages could have conflicting deps, but then we can at least find more of them earlier.

jondequinor:
Rename the script to "pypi_conflicts" or something, and add it to a pipeline so that you don't have to run it manually.

oyvindeide:
> Rename the script to "pypi_conflicts" or something, and add it to a pipeline so that you don't have to run it manually.

The first part I agree on, the second also, but that would be up to the users of `komodo`though. This installs every package, and so is quite slow, running it on every release file would not make sense. The workflow would be on changes to a release file.

oyvindeide:
Something like this looks promising:
https://github.com/tj-actions/changed-files

sondreso:
The snyk job alreadydoes this :slightly_smiling_face:

sondreso:
It transpiles only changed releases: https://github.com/equinor/komodo-releases/blob/master/.github/workflows/snyk.yml#L20-L50

oyvindeide:
Did not think about that - can create something based on that

komodo/check_conflicts.py
```diff
@@ -0,0 +1,130 @@
...
+        type=str,
+        default=False,
+        help="Output file where requirements are written",
+    )
+    parser.add_argument(
+        "--try-pip",
+        action="store_true",
+        help="If given, will try to install packages and check validity. Note that "
+        "creates a virtual environment and is quite slow",
+    )
```
jondequinor, line 62: Do I need to pass `--try-pip` to actually have anything checked? What happens if this is not provided?

jondequinor: Either way, I think invoking this program like so

    komodo-check-conflicts bleeding.yml repository.yml

should

1. if it finds conflict, exit with a non-zero exit code and output these to stdout;
2. if no conflict, exit with 0 without any output.

The parser is called "_Checks if pypi packages are compatible_" and the command is called "**komodo-check-conflicts**", so introducing "output", "requirements", and "try-pip" as default false is a bit unexpected.

Maybe remove `--try-pip`, rename `--output-file` to `--generate-requirements`. The latter should also

1. have help text explaining that the program will "Output installed packages in pip requirements format.";
2. not require a file, but generate the requirements on stdout. If the user needs something in a file, the unix philosophy warrants: 


`komodo-check-conflicts --generate-requirements > requirements.txt`

komodo/check_conflicts.py
```diff
@@ -0,0 +1,130 @@
...
+                )
+            install = subprocess.run(
+                [
+                    current_python,
+                    "-m",
+                    "pip",
+                    "install",
+                    *requirements,
+                    "--no-deps",
+                    "--force-reinstall",  # Probably not needed
```
jondequinor, line 112: not needed

komodo/check_conflicts.py
```diff
@@ -0,0 +1,130 @@
...
+                    current_python,
+                    "-m",
+                    "pip",
+                    "install",
+                    "pip",
+                    "-U",
+                ],
+                stdout=subprocess.PIPE,
+                stderr=subprocess.PIPE,
+            )
```
jondequinor, line 98: Probably equivalent to `capture_output=True`.

oyvindeide: Yes, but that was introduced in `3.7`

komodo/check_conflicts.py
oyvindeide: Yes you do, will remove that option so it always checks.

komodo/check_conflicts.py
jondequinor: Right, thanks for pointing that out!

oysteoh:
Encourage to either 
- Make a comment justifying the PR to exist 
- Close it
- Fix and merge

oyvindeide:
I will have a look at it in the next week or so, if I dont get anywhere I will close it.

PR #233, opened 2021-10-26
Upgrade pip before installing cmake

dafeda:


dafeda:
> But we could automatically grab the latest compatible versions of `pip`, `wheel` and `setuptools` during `komodoenv`'s `setup.py`. So every Komodo release will have mostly up-to-date packages without us having to resort to doing it manually.

[`komodoenv` uses a `project.tml` to specify `requires = ["setuptools>=30.3.0", "wheel", "setuptools_scm"]`.](https://github.com/equinor/komodoenv/blob/master/pyproject.toml).

How do you specify a `pip`-version in `setup.py`?

dotfloat:
You wouldn't. Remember that:
1. Install-time dependencies is different from bundling the dependencies. It's akin to the difference between compiling with a certain library and bundling an installer for the library.
2. Komodo ignores any and all dependencies outside of the ones specified in the release's YAML file
3. We don't want to install `pip` et.al into a Komodo release. That is, `pip` should only be available to those that explicitly used `komodoenv`.

I was thinking something along the lines of:
```py
def fetch_latest_bundle():
    subprocess.run((sys.executable,
                    "-m",
                    "pip",
                    "download",
                    "pip",
                    "wheel",
                    "setuptools",
                    "--dest",
                    path_to_bundle_dir))

fetch_latest_bundle()

setup(
    name="komodoenv",
    ...
)
```

No idea how I'd test it though. (Maybe this is a #yolo situation?)

oysteoh:
Encourage to either 
- Make a comment justifying the PR to exist 
- Close it
- Fix and merge

dafeda:
This does not seem to be terribly important, and I suggest closing it.

PR #238, opened 2021-11-11
Remove logging of komodo usage

ManInFez:
komodo version are logged also when using ERT, thus logging when sourcing does not provide much value.

### DIFF
--- a/komodo/data/enable.csh.in
+++ b/komodo/data/enable.csh.in
@@ -45,24 +45,12 @@ if ( -r $local_script) then
     source $local_script
 endif
 
-alias log_komodo '\\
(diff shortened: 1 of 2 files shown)


oyvindeide (approved):

sondreso:
What about people that use komodo, but not ert?

oyvindeide:
We will not know about them, the main reason for deleting is that the sourcing introduces a lot of noise, so we will probably not know anyway. For example ff someone runs ert locally it will source once per realization, so will add a lot of noise, and everyone opening a terminal will give us one signal, it will be impossible to sort out users who do not use ert. But the threshold to reintroduce this is very small.

PR #239, opened 2021-11-24
Store git hash instead of branch names to release manifest

sondreso:
This will store the full commit hash of the cloned ref to the release
manifest, instead of the tag/branch name. A minimal git repo is
added to the test data to allow for testing of a known hash in
the integration test. This also adds testing of fetching using
git clone in the integration test, which was previously not tested.


jondequinor:
Less ambiguity in this output of komodo is in line with the robustness principle. So 👍🏻.

sondreso:
Tested this with a build, and test-pipelines seems to work as expected with the commit hash in the manifest instead of the branch name :heavy_check_mark:

oysteoh (approved):

PR #242, opened 2022-01-05
Set mpl backend to headless agg

ManInFez:
Newer version of matplotlib is failing at `import matplotlib.pyplot` when running in a headless environment as it tries to set backend to qtagg which requires an interactive environment. The backend can be set by using the env variable MPLBACKEND, and should be headless for all tests running on jenkins. Thus setting it in the testKomodo script.

### DIFF
--- a/vars/testKomodo.groovy
+++ b/vars/testKomodo.groovy
@@ -136,6 +136,7 @@ def call(Map args = [:]) {
                         git log -n 1
                         source ${env.CI_SOURCE_ROOT}/${params.TEST_SCRIPT}
                         git log -n 1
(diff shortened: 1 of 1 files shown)


pinkwah:
Should this be in `jenkins/ci_helper.sh` maybe?

ManInFez:
> Should this be in `jenkins/ci_helper.sh` maybe?

But all those functions are overwritten by the individual testkomodo.sh scripts :shrug:

ManInFez:
the groovy script already sets up a few environment variables, so I thought it made sense to keep them together..

xjules (approved):
LGTM!

PR #243, opened 2022-01-20
Update pysnyk to equinor version enabling latest importlib_metadata

xjules:
Importlib_metadata needs to be recent due to conflicts with upstream pysnyk libraly. Therefore creating our own at the moment until the issue https://github.com/snyk-labs/pysnyk/issues is resolved


pinkwah (approved):
:eyes: :fire:

PR #245, opened 2022-01-20
Drop Python 3.6 support for the tool

pinkwah:
Komodo (the tool) will require Python 3.8. The Python version in the
output Komodo environment is unaffected by this change.


xjules (approved):
LGTM! 🚀

PR #246, opened 2022-01-21
Source python3.8 in symlink suggester

xjules:
Symlink suggester fails due to pysnyk requires `>=Python3.7` hence sourcing Python3.8 first.


oysteoh (approved):
🚀

pinkwah (approved):

PR #247, opened 2022-01-25
Fetch ci_helper using git instead of wget

ManInFez:
Fixes failing wget to githubusers.com

### DIFF
--- a/vars/testKomodo.groovy
+++ b/vars/testKomodo.groovy
@@ -112,7 +112,7 @@ def call(Map args = [:]) {
 
             stage('Run Tests') {
                 steps {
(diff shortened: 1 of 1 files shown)


markusdregi:
Tested it locally and it seems to work 👍🏻 Perhaps this change deserves a comment in the script, the commit message body or both?

ManInFez:
I think maybe we should wait to see if we get the network fixed

ManInFez:
> Tested it locally and it seems to work 👍🏻 Perhaps this change deserves a comment in the script, the commit message body or both?

probably both

sondreso:
I guess we can close this, as the network issues are gone?

PR #253, opened 2022-06-09
Provide newer swig via SWIG_ROOT environment var

jondequinor:
Standard RH7 swig is ancient (2.0.10). A newer swig exists and komodo
should offer to set it, granted that CMP0074[1] has been set to NEW.

[1] https://cmake.org/cmake/help/latest/policy/CMP0074.html


sondreso (approved):
LGTM!

PR #264, opened 2022-10-20
Don't have `testkomodo.sh` automatically run code

pinkwah:
This breaks all current `testkomodo.sh` scripts, but it should make them much simpler in the long run. `testkomodo.sh` scripts can expect `ci_helper.sh` functions to be available with the prefix `ci_`, but other than that it is up to each and every script to handle the setup and testing. `run_tests` is no longer a function that is ran.


vars/testKomodo.groovy
```diff
@@ -137,7 +137,6 @@ def call(Map args = [:]) {
                         source ${env.CI_SOURCE_ROOT}/${params.TEST_SCRIPT}
```
sondreso, line 137: IS the plan that the test script will run whatever it wants to run when sourced?

sondreso (commented):
I think the changes look OK, but when merged, will this not break testing of every current release? :thinking:

jonathan-eq:
Is this still being worked on, or can it be closed? @pinkwah

pinkwah:
Stale.

PR #268, opened 2022-11-08
Use f-strings, fix typo, long options

berland:


komodo/fetch.py
```diff
@@ -60,13 +56,12 @@ def fetch(pkgs, repo, outdir=".", pip="pip"):
 
     if missingpkg:
         eprint("Packages requested, but not found in the repository:")
-        eprint("missingpkg: {}".format(",".join(missingpkg)))
+        eprint("missingpkg: " + ",".join(missingpkg))
```
xjules, line 59: `eprint("missingpkg: {','.join(missingpkg)}")` ?

berland: That is a possibility, is there a reason to prefer one over the other (I don't really know in this case)?

xjules: I thought that `f"{xxx}"` is what we are after :)

berland: I know that `.format()` should be avoided, but do not know a reason to avoid the `+` operator on strings in general.. (but happy to change, just don't know why..)

berland: https://stackoverflow.com/questions/59180574/string-concatenation-with-vs-f-string

xjules: Ok, just keep it as is I would say. It's more like to pursue your `f-formatting` career to the pure perfection :)

xjules (approved):
LGTM!

PR #273, opened 2022-11-25
Drop Python 3.6, add Python 3.10

berland:


berland:
Needs a fix for #274  as well.

berland:
> Needs a fix for #274 as well.

That is fixed by pinning to Ubuntu 20.04.

.github/workflows/test.yml
```diff
@@ -4,14 +4,14 @@ on: [pull_request]
...
     env:
-      RELEASE: release-py37.yml
+      RELEASE: release-py3x.yml
+
 
     strategy:
       fail-fast: false
       matrix:
-        python-version: ["3.6", "3.8"]
+        python-version: ['3.8', '3.10']
```
hnformentin, line 14: How do you think it will impact the users? Have we communicated it beforehand?

berland: We are no longer offering updated Python 3.6 komodo builds, the last available one is 2022.06.

.github/workflows/test.yml
```diff
@@ -21,14 +21,14 @@ jobs:
...
+      uses: actions/setup-python@v4
       with:
         python-version: ${{ matrix.python-version }}
 
     - name: Install dependencies
       run: |
-        sudo apt-get update
-        sudo apt-get install csh
+        sudo apt update
+        sudo apt install csh
```
hnformentin, line 31: Nice, if I understand correctly apt and apt-get have the same behaviour, but apt is more user friendly.

.github/workflows/test.yml
```diff
@@ -44,7 +44,16 @@ jobs:
...
+        ./runkmd.sh \
+            ci/travis/${RELEASE} \
+            ci/travis/repository.yml \
+            --workspace /tmp/kmd-ws \
+            --cache /tmp/kmd-cache \
+            --prefix /tmp/pfx \
+            --release travis \
+            --locations-config $(realpath ci/travis/locations.yml) \
+            --renamer rename.ul
+
```
hnformentin, line 56: Nice, made it more readable.

ci/travis/repository.yml
```diff
@@ -1,12 +1,5 @@
 numpy:
-  1.18.4:
-    source: pypi
-    make: pip
-    maintainer: travis
-    depends:
-      - setuptools
-      - python
-  1.16.6:
+  1.23.5:
```
hnformentin, line 2: Would you know why we had the two versions before and now we have only one?

ci/travis/release-py3x.yml
```diff
@@ -0,0 +1,4 @@
+numpy: 1.23.5
+python: 3-builtin
+setuptools: 65.6.3
+wheel: 0.38.4
```
hnformentin, line 4: Could bumping more recent versions have any unexpected side effect?

hnformentin (approved):
Please take the time to answer the questions.
Unless they trigger any change in the content of this PR, it LGTM.

ci/travis/release-py3x.yml
berland: I don't think so, I think they just serve as an example to test the komodo scripts, that this release matches with the repository etc. But they had to be bumped because it actually pulls these off pypi, and the versions pulled must be supported on the Python version we test for.

ci/travis/repository.yml
berland: No, could not find a reason for it in history. But reducing to one version does not break the tests, so then it should be reduced.

PR #286, opened 2023-01-04
Group kmd arguments and provide better cache default

kwinkunks:
Improves the user interface with new group of 'required named arguments'.
Required overriding the 'optional arguments' group as well to get --help
to continue to make sense.

Also adds a more sensible default value for kmd's --cache (for pip); the
previous default of `.` caused the application to crash.

Not totally sure how this will merge with Håvard's recent PR.

### DIFF
--- a/komodo/build.py
+++ b/komodo/build.py
@@ -326,7 +326,7 @@ def make(
             builddir=builddir,
             makeopts=current.get("makeopts"),
             makefile=current.get("makefile"),
-            dlprefix=dlprefix if dlprefix else ".",
+            dlprefix=dlprefix,
             jobs=jobs,
             cmake=cmk,
             pip=pip,
--- a/komodo/fetch.py
+++ b/komodo/fetch.py
@@ -47,7 +47,7 @@ def grab(path, filename=None, version=None, protocol=None, pip="pip"):
         raise NotImplementedError(f"Unknown protocol {protocol}")
 
 
-def fetch(pkgs, repo, outdir=".", pip="pip"):
+def fetch(pkgs, repo, outdir, pip="pip"):
     missingpkg = [pkg for pkg in pkgs if pkg not in repo]
     missingver = [
         pkg for pkg, ver in pkgs.items() if pkg in repo and ver not in repo[pkg]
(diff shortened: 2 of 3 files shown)


komodo/build.py
```diff
@@ -327,7 +327,7 @@ def resolve(x):
             builddir=builddir,
             makeopts=current.get("makeopts"),
             makefile=current.get("makefile"),
-            dlprefix=dlprefix if dlprefix else ".",
+            dlprefix=dlprefix,
```
berland, line 330: Since this is called  `cache` in `cli.py,` should we rename it here as well?

berland: What happens if someone calls this `make() `now with dlprefix set to an empty string?

berland: (the function signature defaults it to `None`)

kwinkunks: Good point, changing `dlprefix` to `cache` for consistency. 

If the empty string is passed in, then `pushd` in `shell.py` fails when it tries to CD to '' in `fetch.py`. I think it should probably be a positional argument for `make`, and we throw an exception if it's Falsey.

kwinkunks: ...and on further inspection, I'm not so sure. The problem is that the same parameter is called `outdir` in `komodo.fetch.fetch()`, and `output` in the CLI implemented by `komodo.fetch.py`. So there are at least 4 names for the same thing.

I feel like changing the name is going to need a `DeprecationWarning` because who knows who/what is using these various functions. So I think the best approach is probably to leave it as-is and make an issue of the problem.

I have added a trap in `fetch.py` to catch Falsey values though, that bit's ok I think. Tests are passing.

What do you think @berland?

berland: Yes, leaving the API as is can be good for now.

komodo/fetch.py
```diff
@@ -67,7 +67,9 @@ def fetch(pkgs, repo, outdir=".", pip="pip"):
     if missingpkg or missingver:
         return
 
-    if outdir and not os.path.exists(outdir):
+    if not outdir:
+        raise ValueError("The value of `cache` cannot be None or the empty string.")
```
berland, line 71: Is `cache` and not `outdir` intentional here?

PR #287, opened 2023-01-05
add cli tool to query version of package in build

### LINKED ISSUE #278: cli command to get the version of a specific package in komodo release
We should provide an utility to extract a package version used in a specific komodo version. This should be something similar to `pip show <library>` while only extracting the readable version number. The goal is to use it in gh actions workflows to get the proper package version when cloning it.
Eg.
```sh
komodo --show-package ert --release 2022.11.00-py38
>> 4.1.0
```

kwinkunks:
The commit contains a new CLI tool, komodo-show-version (named after pip show, which shows more than just version). It detects the active environment and looks up the version (or git commit hash for non-pip packages) from the 'manifest', a text file komodo creates at build time. Works for komodo and komodoenv environments.

The commit also contains tests.

fixes #278

### DIFF
--- a/komodo/show_version.py
+++ b/komodo/show_version.py
@@ -1,0 +1,181 @@
+#!/usr/bin/env python
+import argparse
+import configparser
+import os
+import re
+import sys
+import textwrap
+from pathlib import Path
+from typing import Dict, List, Optional, Union
+
+import yaml
+
+from komodo.yaml_file_type import ManifestFile
+
+
+def get_release() -> str:
+    """
+    Get the komodo release from the active environment.
+
+    The value of the KOMODO_RELEASE environment variable implicitly tells
+    us the kind of environment we are in. If it looks like an ordinary name,
+    it is probably a komodo environment. However, if it is a path, it is a
+    komodoenv environment.
+
+    We need to know the kind of environment because komodo and komodoenv
+    environments store their manifest files (and root folders) in different
+    places.
+    """
+    try:
+        return os.environ["KOMODO_RELEASE"]
+    except KeyError:
+        message = textwrap.dedent(
+            """
+            No active komodo environment found.
+
+            Either run this command from an active komodo or komodoenv environment,
+            or provide an environment manifest file path with the --path option.\
+            """
+        )
+        sys.exit(message)
+
+
+def read_config(fname: Union[str, Path]) -> dict:
+    """
+    Read an INI file (aka config file) that does not have sections.
+    Sections are part of the INI file format specification and cannot be read
+    by a `configparser.ConfigParser` without them. This function creates a
+    temporary 'MAIN' section and passes back its contents.
+
+    Args:
+        fname: the path to a 'flat' config file, i.e. one with no sections.
+
+    Returns:
+        The configuration.
+    """
+    config = configparser.ConfigParser()
+    with open(fname, "rt", encoding="utf-8") as stream:
+        config.read_string("[MAIN]\n" + stream.read())
+    return dict(config["MAIN"])
+
+
+def get_komodoenv_path(release: str) -> Path:
+    """
+    Use the release name to find the 'real' release path, but from a
+    komodoenv environment. These environments overwrite the environment
+    variable $KOMODO_RELEASE and $PATH, but they store the original path
+    in the komodoenv at 'root/pyvenv.cfg'.
+
+    Args:
+        release: The name of the release, e.g. as stored in the KOMODO_RELEASE
+            environment variable.
+
+    Returns:
+        The path to the release, where the 'root/bin' folder is.
+    """
+    config = read_config(Path(release) / "root" / "pyvenv.cfg")
+    path, *_ = config["home"].split("/root/bin")
+    return Path(path)
+
+
+def get_komodo_path(release: str) -> Path:
+    """
+    Use the release name to find the 'real' release path in an ordinary
+    komodo environment. E.g., the release may be something like
+    '2023.01.02-py38' but the 'real' release (where the release manifest
+    is stored) might be platform-specific, e.g. '2023.01.02-py38-rhel7'.
+    The real path is in the PATH, so we try to get it from there.
+
+    Args:
+        release: The name of the release, e.g. from $KOMODO_RELEASE.
+
+    Returns:
+        The path to the release, where the 'root/bin' folder is.
+    """
+    paths = os.environ["PATH"]
+    pattern = re.compile(rf"(.*?{release}.*?)/root/bin")
+    result = pattern.search(paths)
+    if result is not None:
+        (path,) = result.groups()
+    else:
+        message = f"Could not retrieve the path to the release {release}."
+        raise RuntimeError(message)
+    return Path(path)
+
+
+def get_version(pkg: str, manifest: Optional[Dict] = None) -> str:
+    """
+    Get the release number (or git commit hash) for a package in a
+    komodo release file. If no file is specified, the current release
+    is used. If no environment is active, the path to the release
+    must be specified.
+
+    Args:
+        pkg: The name of the package to get the version for.
+        manifest: Optional. A mapping of packages to dicts of version and
+            maintainer. Typically, this will come from the YAML file created
+            by komodo when it builds a release.
+
+    Returns:
+        The version number or git hash of the version.
+    """
+    path = None
+    if manifest is None:
+        release = get_release()
+        if (Path(release) / "komodoenv.conf").is_file():
+            path = get_komodoenv_path(release)
+        else:
+            path = get_komodo_path(release)
+        release_file = path.parts[-1]
+
+        with open(path / release_file, "rt", encoding="utf-8") as stream:
+            manifest = yaml.safe_load(stream)
+
+    package = manifest.get(pkg)
+
+    if package is None:
+        path_str = f" {path / release_file}" if path is not None else ""
+        message = f"The package {pkg} is not found in the manifest file{path_str}."
+        raise KeyError(message)
+
+    return package.get("version")
+
+
+def parse_args(args: List[str]) -> argparse.Namespace:
+    """
+    Parse the arguments from the command line into an `argparse.Namespace`.
+    Having a separated function makes it easier to test the CLI.
+
+    Args:
+        args: A sequence of arguments, e.g. as collected from the command line.
+
+    Returns:
+        The `argparse.Namespace`, a mapping of arg names to values.
+    """
+    parser = argparse.ArgumentParser(
+        description="Return the version of a specified package in the active "
+        "release or in a given release manifest file."
+    )
+    parser.add_argument("package", help="Package to find the version for.")
+    parser.add_argument(
+        "--manifest-file",
+        type=ManifestFile(),
+        required=False,
+        help="The full path to a release manifest file. This file is "
+        "produced by komodo when it builds an environment. If omitted, "
+        "komodo-show-version will try to find the file for the active "
+        "environment.",
+    )
+    return parser.parse_args(args)
+
+
+def main() -> str:
+    """
+    Run the CLI and return the result from get_version().
+    """
+    args = parse_args(sys.argv[1:])
+    return get_version(args.package, manifest=args.manifest_file)
+
+
+if __name__ == "__main__":
+    print(main())
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1,0 +1,89 @@
+import os
+from unittest.mock import mock_open, patch
+
+import pytest
+
+
+@pytest.fixture()
+def mock_komodo_env_vars():
+    """
+    Provide the environment vars from a komodo environment.
+    """
+    env = {
+        "PATH": "/foo/bar/komodo-release-0.0.1/root/bin",
+        "KOMODO_RELEASE": "komodo-release-0.0.1",
+    }
+    with patch.dict(os.environ, env):
+        yield
+
+
+@pytest.fixture()
+def mock_bad_komodo_env_vars():
+    """
+    Provide environment vars from a komodo environment, but the variables are
+    not consistent with each other.
+    """
+    env = {
+        "PATH": "/foo/bar/komodo-release-99.99.99/root/bin",
+        "KOMODO_RELEASE": "komodo-release-0.0.1",
+    }
+    with patch.dict(os.environ, env):
+        yield
+
+
+@pytest.fixture()
+def mock_komodoenv_env_vars():
+    """
+    Provide the environment vars for a *komodoenv* environment. These are
+    different from komodo environments in that the name will not necessarily
+    match an original komodo release, and the given path does not contain a
+    komodo manifest file.
+    """
+    env = {
+        "PATH": "/quux/komodo-release/root/bin",
+        "KOMODO_RELEASE": "/quux/komodo-release",
+    }
+    with patch.dict(os.environ, env):
+        yield
+
+
+@pytest.fixture()
+def mock_config_file():
+    """
+    Mock a sectionless INI file, which is the format komodoenv writes into
+    the pyvenv.cfg file that points to the real komodo root.
+    """
+    read_data = (
+        "home = /foo/bar/komodo-release-0.0.1/root/bin\n"
+        "include-system-site-packages = false\n"
+        "version = 3.8.14\n"
+    )
+    m = mock_open(read_data=read_data)
+    with patch("builtins.open", m):
+        yield m
+
+
+@pytest.fixture()
+def mock_version_manifest():
+    """
+    Mock the manifest file that komodo creates when building an environment.
+    """
+    read_data = (
+        "foo:\n  maintainer: jdoe\n  version: 1.2.3\n"
+        "bar:\n  maintainer: jbloggs\n  version: 99.99.99"
+    )
+    m = mock_open(read_data=read_data)
+    with patch("builtins.open", m):
+        yield m
+
+
+@pytest.fixture()
+def mock_release_file():
+    """
+    Mock a release specification file that komodo _consumes_ when building
+    an environment. Passing it to komodo-show-version raises an error.
+    """
+    read_data = "foo: 1.2.3\nbar: 99.99.99"
+    m = mock_open(read_data=read_data)
+    with patch("builtins.open", m):
+        yield m
--- a/tests/test_show_version.py
+++ b/tests/test_show_version.py
@@ -1,0 +1,145 @@
+from pathlib import Path
+
+import pytest
+
+from komodo.show_version import (
+    get_komodo_path,
+    get_komodoenv_path,
+    get_release,
+    get_version,
+    parse_args,
+    read_config,
+)
+
+
+def test_read_config_file(mock_config_file):
+    """Test the function that reads sectionless INI files."""
+    config = read_config("foo.cfg")
+    mock_config_file.assert_called_once_with("foo.cfg", "rt", encoding="utf-8")
+    assert config["home"] == "/foo/bar/komodo-release-0.0.1/root/bin"
+
+
+def test_get_release(mock_komodo_env_vars):
+    """Check the environment variable is picked up."""
+    assert get_release() == "komodo-release-0.0.1"
+
+
+def test_get_release_fails_without_env_var():
+    """
+    Check that the script exists gracefully if not run in a komodo environment.
+    """
+    with pytest.raises(SystemExit) as exception_info:
+        _ = get_release()
+    assert "No active komodo environment found." in str(exception_info.value)
+
+
+def test_get_komodo_path(mock_komodo_env_vars):
+    """
+    This is the standard scenario when using an ordinary komodo (not komodoenv)
+    environment. This test checks that the correct path is detected from the
+    mocked environment.
+    """
+    release = get_release()
+    path = Path("/foo/bar/komodo-release-0.0.1")
+    assert get_komodo_path(release) == path
+
+
+def test_get_komodoenv_path(mock_komodoenv_env_vars, mock_config_file):
+    """
+    Komodoenv environments wrap komodo environments and hide the location of
+    the manifest file. Test that the `get_komodoenv_path()` function recovers
+    the actual manifest file path from the `pyvenv.cfg` file.
+    """
+    release = get_release()
+    path = Path("/foo/bar/komodo-release-0.0.1")
+    assert get_komodoenv_path(release) == path
+    fname = Path("/quux/komodo-release/root/pyvenv.cfg")
+    mock_config_file.assert_called_once_with(fname, "rt", encoding="utf-8")
+
+
+def test_get_version(mock_komodo_env_vars, mock_version_manifest):
+    """
+    This test should pick up the $KOMODO_RELEASE then correctly
+    read the mocked release 'manifest' file.
+    """
+    fname = Path("/foo/bar/komodo-release-0.0.1/komodo-release-0.0.1")
+    assert get_version("foo") == "1.2.3"
+    mock_version_manifest.assert_called_once_with(fname, "rt", encoding="utf-8")
+
+
+def test_get_version_with_filepath(mock_version_manifest):
+    """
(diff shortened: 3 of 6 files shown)


komodo/show_version.py
```diff
@@ -0,0 +1,133 @@
...
+        raise Exception("Could not sense the name of the release, please provide it.")
+
+
+def read_config(fname: Union[str, Path]) -> dict:
+    """
+    Read an INI file (aka config file) that does not have sections.
+    Creates a temporary 'MAIN' section and pass bask its contents.
+    """
+    config = configparser.ConfigParser()
+    with open(fname, "rt") as f:
```
berland, line 28: add `encoding="utf-8"`

komodo/show_version.py
```diff
@@ -0,0 +1,133 @@
...
+
+def guess_komodoenv_path(release: str) -> Path:
+    """
+    Use the release name to find the 'real' release path, but from a
+    komodoenv environment. These environments overwrite the environment
+    variable $KOMODO_RELEASE and $PATH, but they store the original path
+    in the komodoenv at 'root/pyvenv.cfg'.
+
+    Args:
+        release (str): The name of the release, e.g. from $KOMODO_RELEASE.
```
berland, line 41: We don't need to specify the type inside the docstrings, it is (or can be) inferred from the function signature.

komodo/show_version.py
```diff
@@ -0,0 +1,133 @@
...
+
+    Returns:
+        str: The path to the release, where the 'root/bin' folder is.
+    """
+    config = read_config(Path(release) / "root" / "pyvenv.cfg")
+    path, *_ = config["home"].split("/root/bin")
```
berland, line 47: (`maxsplit=1` is an alternative to `*_`)

tests/test_show_version.py
```diff
@@ -0,0 +1,111 @@
...
+
+
+@pytest.fixture()
+def mock_komodoenv_env_vars():
+    """
+    Provide the 'fake' environment vars from a komodoenv environment.
+    """
+    env = {
+        "PATH": "/quux/komodo-release/root/bin",
+        "KOMODO_RELEASE": "/quux/komodo-release",
```
berland, line 36: I suppose this test covers the case in https://github.com/equinor/komodo-releases/pull/3043/files#diff-15f0a4593f5be26df91be4cafda7792e49d2c795c2b82ef1e72b58bd8abf89b1R30-R31

komodo/show_version.py
```diff
@@ -0,0 +1,133 @@
...
+        else:
+            path = guess_komodo_path(release)
+        release_file = path.parts[-1]
+
+    with open(path / release_file, "rt") as f:
+        packages = yaml.safe_load(f)
+    package = packages.get(pkg)
+
+    if package is None:
+        raise ValueError(f"The package {pkg} is not in the release.")
```
berland, line 106: Add the content of the `release` string to the error message?

komodo/show_version.py
```diff
@@ -0,0 +1,133 @@
...
+        path = path.parent
+    else:
+        release = guess_release()
+        if (Path(release) / "komodoenv.conf").is_file():
+            path = guess_komodoenv_path(release)
+        else:
+            path = guess_komodo_path(release)
+        release_file = path.parts[-1]
+
+    with open(path / release_file, "rt") as f:
```
berland, line 101: (Pylint will complain about these one-letter variable names)

tests/test_show_version.py
```diff
@@ -0,0 +1,111 @@
...
+def test_guess_komodo_path(mock_komodo_env_vars):
+    release = guess_release()
+    path = pathlib.Path("/foo/bar/komodo-release-0.0.1")
+    assert guess_komodo_path(release) == path
+
+
+def test_guess_komodoenv_path(mock_komodoenv_env_vars, mock_config_file):
+    """
+    Even though the release and PATH point to 'fake' environments, this
+    shold recover the true path from the pyvenv.cfg file.
```
berland, line 95: shold -> should

komodo/show_version.py
```diff
@@ -0,0 +1,133 @@
...
+
+
+def guess_release() -> str:
+    """
+    Try to sense the Komodo release from the active environment.
+    """
+    try:
+        return os.environ["KOMODO_RELEASE"]
+    except KeyError:
+        raise Exception("Could not sense the name of the release, please provide it.")
```
berland, line 19: We should not use bare exceptions. You may use a dummy exception `KomodoShowVersionException` that just inherits from `Exception` if needed.

berland:
Is this supposed to work or should it fail like this?
```
 [havb@be-lx895926:/work/projects/komodo] cli-get-version$ komodo-show-version  --path /prog/res/komodo/2022.10.03-py38-rhel7/2022.10.03-py38-rhel7  ecl2df
Traceback (most recent call last):
  File "/work/venv/new/bin/komodo-show-version", line 33, in <module>
    sys.exit(load_entry_point('Komodo', 'console_scripts', 'komodo-show-version')())
  File "/work/projects/komodo/komodo/show_version.py", line 129, in main
    return get_version(args.package, args.path)
  File "/work/projects/komodo/komodo/show_version.py", line 101, in get_version
    with open(path / release_file, "rt") as f:
FileNotFoundError: [Errno 2] No such file or directory: '/prog/res/komodo/2022.10.03-py38-rhel7/2022.10'
```

komodo/show_version.py
kwinkunks: Fair point, esp since two of them were wrong :grimacing:

komodo/show_version.py
kwinkunks: Not sure, still requires at least the comma (i.e. `path, = str.split(maxsplit=1)`) because the split always returns at least two items. Will keep current unless you really don't like that syntax.

komodo/show_version.py
```diff
@@ -0,0 +1,133 @@
...
+def get_version(pkg: str, fpath: Optional[str] = None) -> str:
+    """
+    Get the release number (or git commit hash) for a package in a
+    komodo release file. If no file is specified, the current release
+    is used. If no environment is active, the path to the release
+    must be specified.
+
+    Args:
+        pkg (str): The name of the package to get the version for.
+        fpath (str): Optional. The path to a release 'manifest' file.
```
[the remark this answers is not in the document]
kwinkunks: Changed. I'm never sure what to call file paths. I want to be explicit that it wants the **full path** to the 'manifest' file (and name I made up, by the way, but 'release file' is something else).

komodo/show_version.py
kwinkunks: Good one.

komodo/show_version.py
kwinkunks: Changed to `RuntimeError`, what do you think? Definitely an improvement, not sure what's best.

tests/test_show_version.py
kwinkunks: fxed

tests/test_show_version.py
kwinkunks: :thinking: I'll ask you more about this when I see you.

komodo/show_version.py
```diff
@@ -0,0 +1,133 @@
...
+    )
+    parser.add_argument("package", help="Package to find version for.")
+    parser.add_argument(
+        "--path",
+        required=False,
+        help="The full path to a release file. By default, komodo will try "
+        "to find the file for the active environment.",
+    )
+    args = parser.parse_args()
+    return get_version(args.package, args.path)
```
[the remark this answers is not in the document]
kwinkunks: Good point, I'll think about how to avoid that.

kwinkunks:
Will review this one too, thank you Håvard!

komodo/show_version.py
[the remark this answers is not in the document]
kwinkunks: I think the main situation when we know what's wrong is when there's no $KOMODO_RELEASE. If that was found and the program continues, everything else "should work" and is an exception. So I'll catch that scenario (`KeyError` for the env lookup) and call `sys.exit` in that case.

(`argparse` already deals with mangled/missing args)

Sound reasonable, or did I get the wrong end of the stick @berland ?

komodo/show_version.py
berland: ah, then keep it :)

komodo/show_version.py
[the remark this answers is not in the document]
berland: Sounds good!

tests/test_show_version.py
berland: I was specifically looking for a test that covers the case when the env variable `$KOMODO_RELEASE` contains the full path and not just a string, since I used that as a workaround in the link above. And it looks like you have covered it 👍🏻

kwinkunks:
Of course, I wasn't testing passing in the `path` argument, so I earned this last bug with my negligence. Fixed. (And testing for it now :)

tests/test_show_version.py
kwinkunks: I think it does.

kwinkunks:
Decided to defer the refactoring to `click` (or maybe `rich-cli`?) and PR this again as-is. I think it still needs approval from someone. Cc @berland @xjules

berland:
This one I would think should work:
```
$  komodo-show-version --path examples/releases/ecl.yml libecl
<snip>
AttributeError: 'str' object has no attribute 'get'
```

berland:
It probably wouldn't hurt to use what's in  `examples` in tests, also to ensure the examples does not get too outdated..

kwinkunks:
`komodo-show-version --path examples/releases/ecl.yml libecl` should actually not work currently, because that's a release 'specification' file, whereas the so-called 'manifest' files I'm expecting are created by `komodo` when it builds the release. It _should_ match the specification, of course... but I think the idea is that it's a less canonical source of the version information.

But it raises the question what this tool should do if someone gives it the specification file. Options are:

- Error (more informatively than I am currently erroring) <- I think this is my preference, but could be convinced otherwise
- Return the version but with a warning because we don't quite trust this source
- Just return the version because we trust this source as much as the manifest

This is also the reason I didn't use `examples` in testing --- there's no example of a manifest file currently, so the test would have to build a komodo release first. Do you think that would be a better approach?

berland:
> But it raises the question what this tool should do if someone gives it the specification file. Options are:
> 
> * Error (more informatively than I am currently erroring) <- I think this is my preference, but could be convinced otherwise

I would go for this, maybe just state that the yaml file has a wrong format, optinally saying the format should be like this and that (then the future developer might deduce which of these files it is meant to work on)

tests/test_show_version.py
```diff
@@ -0,0 +1,185 @@
...
+def test_guess_komodo_path_fails(mock_bad_komodo_env_vars):
+    """
+    If the environment has inconsistent information, or the release name
+    does not match the path, then an exception is raised.
+    """
+    release = guess_release()
+    assert release == "komodo-release-0.0.1"
+    with pytest.raises(RuntimeError) as excinfo:
+        _ = guess_komodo_path(release)
+    assert "Could not sense the path to the release" in str(excinfo.value)
```
xjules, line 185: sense -> retrieve?

kwinkunks: Agreed :heavy_check_mark:

kwinkunks:
Moved fixtures to `conftest.py` after review from @xjules, see #303.

kwinkunks:
Cannot see how to get `pytest` to see the 'outer' exception in `komodo.show_version.get_version()`. It only wants to see the 'inner' version. This doesn't affect the behaviour of the tool, but it does mean `test_show_version.test_get_version_fails_with_release_file()` is only checking the 'inner' exception.

komodo/show_version.py
```diff
@@ -0,0 +1,154 @@
...
+    if package is None:
+        raise KeyError(f"The package {pkg} is not found in the release.")
+
+    return package.get("version")
+
+
+def parse_args(args: List[str]) -> argparse.Namespace:
+    parser = argparse.ArgumentParser(
+        description="Return the version of a specified package in a given "
+        "release or release 'manifest' file."
```
hnformentin, line 130: Why does `manifast` need to be in ''?

tests/conftest.py
```diff
@@ -0,0 +1,85 @@
+import os
+from unittest.mock import mock_open, patch
+
+import pytest
+
+
+@pytest.fixture()
+def mock_komodo_env_vars():
+    """
+    Provide the environment vars from a 'vanilla' komodo environment.
```
hnformentin, line 10: Why vanilla with '' here and below? Same with 'fake'...

tests/test_show_version.py
```diff
@@ -0,0 +1,127 @@
+import pathlib
```
hnformentin, line 1: What about `from pathlib import Path`?

tests/test_show_version.py
```diff
@@ -0,0 +1,127 @@
...
+    Check that the script exists gracefully if not run in a komodo environment.
+    """
+    with pytest.raises(SystemExit) as excinfo:
+        _ = guess_release()
+    assert "No active komodo environment found." in str(excinfo.value)
+
+
+def test_guess_komodo_path(mock_komodo_env_vars):
+    """
+    This is the 'standard' scenario when using a vanilla komodo (not komodoenv)
```
hnformentin, line 38: In this file also there are some uses of '' that seems to be unnecessary.

tests/test_show_version.py
```diff
@@ -0,0 +1,127 @@
...
+def test_guess_release(mock_komodo_env_vars):
+    """Simply reads an environment variable."""
+    assert guess_release() == "komodo-release-0.0.1"
+
+
+def test_guess_release_fails_without_env_var():
+    """
+    Check that the script exists gracefully if not run in a komodo environment.
+    """
+    with pytest.raises(SystemExit) as excinfo:
```
hnformentin, line 31: `excinfo` is very difficult to pronounce. What do you think about refactoring this name?

hnformentin (commented):
Proactive use of docstrings, which is something we are looking for to improve in the team. We had a discussion some time ago where we stressed (1) the use of docstrings to explain why instead of what, and (2) the right granularity, which also implies not to write trivial documentation. More can be found [here](https://github.com/equinor/scout/blob/cf8de3f83b88e262ffcb3162d20bac8c4269392e/testing-and-documentation-guide.md#documentation), if you have time to consider for this PR.

komodo/show_version.py
```diff
@@ -0,0 +1,154 @@
+#!/usr/bin/env python
+import argparse
+import configparser
+import os
```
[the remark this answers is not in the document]
berland: I tend to prefer not to import individual functions, as it makes it more readable when that function is in use further down in the code (then I don't have to look at the top to see where that function comes from). But that might come at a speed penalty?

komodo/show_version.py
```diff
@@ -0,0 +1,154 @@
...
+    places.
+    """
+    try:
+        return os.environ["KOMODO_RELEASE"]
+    except KeyError:
+        sys.exit(
+            "\nNo active komodo environment found.\n\nEither run this command "
+            "from an active komodo or komodoenv environment,\nor provide an "
+            "environment 'manifest' file path with the --path option."
+        )
```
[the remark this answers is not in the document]
berland: I see sometimes this is solved by use of `textwrap.dedent`. Random example: https://github.com/equinor/ert/blob/898ce11efb4d749aea8aba18ab02386486eb201e/tests/unit_tests/c_wrappers/res/enkf/test_run_path_creation.py#L22

komodo/show_version.py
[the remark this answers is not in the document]
hnformentin: That is indeed more readable to have the full path. 
I agree that the downside can be speed. Do you think this would make any difference in terms of security? 

There is a way to import one individual function and keep readability (from python documentation):
"""
Users of the package can import individual modules from the package, for example:

import sound.effects.echo
This loads the submodule sound.effects.echo. It must be referenced with its full name.

sound.effects.echo.echofilter(input, output, delay=0.7, atten=4)
"""

How does it look for you?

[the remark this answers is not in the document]
kwinkunks: I'm never sure which way to go with this, but in general I think I prefer more namespace. I don't think it's a speed thing: the entire module is evaluated either way.

It's true you can import submodules with that other pattern, so `import os.path` would be okay, but there's no other way to import `argparse.ArgumentParser` for example. I'm not using any submodules so I'm not sure there's much to do here.

komodo/show_version.py
```diff
@@ -0,0 +1,154 @@
...
+
+def guess_komodoenv_path(release: str) -> Path:
+    """
+    Use the release name to find the 'real' release path, but from a
+    komodoenv environment. These environments overwrite the environment
+    variable $KOMODO_RELEASE and $PATH, but they store the original path
+    in the komodoenv at 'root/pyvenv.cfg'.
+
+    Args:
+        release: The name of the release, e.g. from $KOMODO_RELEASE.
```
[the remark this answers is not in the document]
kwinkunks: Oh right, I meant the environment variable; edited to make it clearer.

komodo/show_version.py
```diff
@@ -0,0 +1,154 @@
...
+            path = guess_komodo_path(release)
+        release_file = path.parts[-1]
+
+        with open(path / release_file, "rt", encoding="utf-8") as stream:
+            manifest = yaml.safe_load(stream)
+
+    package = manifest.get(pkg)
+
+    if package is None:
+        raise KeyError(f"The package {pkg} is not found in the release.")
```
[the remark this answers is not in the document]
kwinkunks: I did have it like that, but now that I'm parsing the file using `argparse`, this function does not always know the path.

Anyway, I've added the path, if available. But it feels a bit polymorphic...

tests/test_show_version.py
kwinkunks: :+1:

tests/test_show_version.py
kwinkunks: I guess this is me continuing to struggle with emphasizing that there are two different scenarios. Maybe I should move the `komodoenv` tests to another file, that might help.

kwinkunks: (There are 3 scenarios though, because passing in a manifest file is another option. And I don't think we want 3 test files.)

tests/test_show_version.py
kwinkunks: It's the name they use in the `pytest` docs, but I'm happy to change it. `exception_info`?

komodo/show_version.py
```diff
@@ -0,0 +1,154 @@
...
+from typing import Dict, List, Optional, Union
+
+import yaml
+
+from komodo.yaml_file_type import ManifestFile
+
+
+def guess_release() -> str:
+    """
+    Try to sense the Komodo release from the active environment.
```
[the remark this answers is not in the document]
kwinkunks: :+1:

komodo/show_version.py
```diff
@@ -0,0 +1,154 @@
...
+
+    The value of the KOMODO_RELEASE environment variable implicitly tells
+    use the kind of environment we are in. If it looks like an ordinary name,
```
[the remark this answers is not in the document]
kwinkunks: Changed to "tells us".

komodo/show_version.py
```diff
@@ -0,0 +1,154 @@
...
+
+def read_config(fname: Union[str, Path]) -> dict:
+    """
+    Read an INI file (aka config file) that does not have sections.
+    Creates a temporary 'MAIN' section and passes back its contents.
+    """
+    config = configparser.ConfigParser()
+    with open(fname, "rt", encoding="utf-8") as stream:
+        config.read_string("[MAIN]\n" + stream.read())
+    return dict(config["MAIN"])
```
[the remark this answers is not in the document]
kwinkunks: Done, added to docstring.

tests/test_show_version.py
kwinkunks: Done

tests/conftest.py
kwinkunks: Again, tidied up with more explicit language.

komodo/show_version.py
```diff
@@ -0,0 +1,154 @@
...
+    """
+    Get the release number (or git commit hash) for a package in a
+    komodo release file. If no file is specified, the current release
+    is used. If no environment is active, the path to the release
+    must be specified.
+
+    Args:
+        pkg: The name of the package to get the version for.
+        manifest: Optional. A 'manifest' mapping packages to dicts
+            of version and maintainer.
```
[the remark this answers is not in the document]
kwinkunks: This indentation is standard for Google-style docstrings (this package does not have a standard yet though). I actually assumed `black` would do whatever it wants but it seems not to touch docstrings?

komodo/show_version.py
```diff
@@ -0,0 +1,154 @@
...
+            "\nNo active komodo environment found.\n\nEither run this command "
+            "from an active komodo or komodoenv environment,\nor provide an "
+            "environment 'manifest' file path with the --path option."
+        )
+
+
+def read_config(fname: Union[str, Path]) -> dict:
+    """
+    Read an INI file (aka config file) that does not have sections.
+    Creates a temporary 'MAIN' section and passes back its contents.
```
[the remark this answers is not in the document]
kwinkunks: Reworded to deal with explanation of why this function exists.

komodo/show_version.py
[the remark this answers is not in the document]
kwinkunks: Lol, flake8 fails with W293 (blank line has whitespace) if you have a blank line in the middle of a string literal.

kwinkunks:
Thanks for the reviews, all!

komodo/show_version.py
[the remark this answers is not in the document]
hnformentin: Thanks! There is still one extra \ by the end of the message.

komodo/show_version.py
```diff
@@ -0,0 +1,180 @@
...
+        release_file = path.parts[-1]
+
+        with open(path / release_file, "rt", encoding="utf-8") as stream:
+            manifest = yaml.safe_load(stream)
+
+    package = manifest.get(pkg)
+
+    if package is None:
+        path_str = f" {path / release_file}" if path is not None else ""
+        message = f"The package {pkg} is not found in the manifest file{path_str}."
```
hnformentin, line 137: Missing a blank space in `file{path_str}`, correct?

komodo/show_version.py
hnformentin: I understand. In this case, what do you think about adding this nomenclature in the [release manager manual](https://github.com/equinor/scout/blob/2e2086ec59fae3492f098b8114f1e83d9f79fcf2/release-process/README.md)? If you think it is a good idea, just creating the issue would be suffice since solving it is not exactly on the scope of this PR.

tests/conftest.py
hnformentin: Your explanation here is really clear and indeed, this difference is not evident. 
I think they could be arranged in the docstrings of `mock_komodo_env_vars` and `mock_komodoenv_env_vars`, for example, and that would probably be useful for a future developer.

tests/test_show_version.py
hnformentin: That I did not know, thanks for considering.

kwinkunks:
Just realized I did not add the tool to the README. Done, with a small clarification to the CLI help string as well.

tests/conftest.py
kwinkunks: Added some clarification to `conftest.py`.

komodo/show_version.py
kwinkunks: done

komodo/show_version.py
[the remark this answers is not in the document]
kwinkunks: Right... I was trying to sidestep a `pylint` issue, but in a stupid way. Fixed.

komodo/show_version.py
kwinkunks: It does look weird but it's correct, because sometimes the sentence ends there.

hnformentin (approved):
Looks good, appreciate considering all the comments :-)

Issue #289, opened 2023-01-05
Pylint komodo


berland:
* Pick (one or more) pylint issue from the list and make a PR for it
* Add solved pylint issues to CI when fixed to avoid reintroduction.

```
$ pylint --reports=y --enable=all komodo tests | tee pylintreport.txt
[...]


Messages by category
--------------------

+-----------+-------+---------+-----------+
|type       |number |previous |difference |
+===========+=======+=========+===========+
|convention |416    |154      |154        |
+-----------+-------+---------+-----------+
|refactor   |150    |200      |200        |
+-----------+-------+---------+-----------+
|warning    |22     |389      |389        |
+-----------+-------+---------+-----------+
|error      |0      |20       |20         |
+-----------+-------+---------+-----------+



Messages
--------

+---------------------------------------+------------+
|message id                             |occurrences |
+=======================================+============+
|missing-function-docstring             |314         |
+---------------------------------------+------------+
|duplicate-code                         |104         |
+---------------------------------------+------------+
|missing-module-docstring               |64          |
+---------------------------------------+------------+
|line-too-long                          |18          |
+---------------------------------------+------------+
|too-many-arguments                     |16          |
+---------------------------------------+------------+
|redefined-outer-name                   |13          |
+---------------------------------------+------------+
|missing-class-docstring                |12          |
+---------------------------------------+------------+
|too-many-locals                        |9           |
+---------------------------------------+------------+
|no-else-return                         |8           |
+---------------------------------------+------------+
|protected-access                       |6           |
+---------------------------------------+------------+
|use-implicit-booleaness-not-comparison |5           |
+---------------------------------------+------------+
|too-few-public-methods                 |5           |
+---------------------------------------+------------+
|suppressed-message                     |4           |
+---------------------------------------+------------+
|locally-disabled                       |4           |
+---------------------------------------+------------+
|too-many-statements                    |3           |
+---------------------------------------+------------+
|too-many-branches                      |3           |
+---------------------------------------+------------+
|consider-using-dict-items              |3           |
+---------------------------------------+------------+
|unnecessary-lambda                     |2           |
+---------------------------------------+------------+
|consider-using-with                    |2           |
+---------------------------------------+------------+
|logging-fstring-interpolation          |1           |
+---------------------------------------+------------+




------------------------------------------------------------------
Your code has been rated at 8.61/10 (previous run: 9.77/10, -1.16)
```

eivindjahren:
[...]

PR #290, opened 2023-01-06
Add pylint to CI

berland:
Currently every current pylint issue is exempted.

Excemptions (disables) in .pylintrc are to be removed one-by-one in upcoming PRs.


berland:
See #289

.github/workflows/style.yml
```diff
@@ -39,3 +40,7 @@ jobs:
     - name: Lint with flake8
```
kwinkunks, line 40: Just a question: is the idea that `pylint` will _replace_ `flake8`, maybe when its errors have been dealt with? Or do we normally have both on a project?

.pylintrc
```diff
@@ -0,0 +1,63 @@
+[MESSAGES CONTROL]
```
kwinkunks, line 1: Instead of having a dot file to control `pylint`, it might be preferable to put its options in `setup.cfg` (and, eventually, `pyproject.toml` if/when we move to PEP621-style packaging).

The `disable` list would go into a `[pylint.messages control]` section of `setup.cfg`.

.github/workflows/style.yml
berland: We will run both.

.pylintrc
berland: Thanks, that made it prettier. Used `setup.cfg`.

.github/workflows/style.yml
```diff
@@ -27,6 +27,7 @@ jobs:
 
     - name: Install dependencies
       run: |
+        pip install .
```
berland, line 30: @xjules , it seems your comment here has been lost. This line is added in order to let `pylint` check that everything is importable, so this is used only to get the dependencies in, not to get komodo itself in.

xjules (approved):
:rocket: :fork_and_knife:

PR #297, opened 2023-01-09
Solve some pylint issues

berland:
Based off #290


lars-petter-hauge:
Could you squash the two commits for `Add pylint to ci`? (and technically they should have the inverse order, otherwise the commit introducing the pylint would be a failing commit)

.github/workflows/style.yml
```diff
@@ -27,6 +27,7 @@ jobs:
 
     - name: Install dependencies
       run: |
+        pip install .
```
xjules, line 30: Do we need to install komodo too for style checking?

berland: yes, there was a reason, need to rething what it was..

berland: It is in order to obtain the dependencies

berland:
> Could you squash the two commits for `Add pylint to ci`? (and technically they should have the inverse order, otherwise the commit introducing the pylint would be a failing commit)

Fixed!

komodo/reverse_dep_graph.py
```diff
@@ -58,13 +58,13 @@ def _dump_dot(reverse, pkg, version, out):
 def _dump_dot_dep(reverse, pkg, version, out, seen):
     if pkg in seen:
         return
-    id = pkg.lower().replace("-", "_")
-    out.write(f'  {id} [label="{pkg}-{version}"];\n')
+    _id = pkg.lower().replace("-", "_")
```
kwinkunks, line 61: I think it's more conventional to use a _trailing_ underscore to avoid a collision with a builtin. That's what PEP8 says at least. (The leading underscore signals 'internal use' --- which this is too, I guess.)

kwinkunks:
Other than the `_id`, which is not a big deal at all, I think everything in here looks good. Most of it is a clear improvement, and I learned some things about `pylint` too.

komodo/reverse_dep_graph.py
berland: fixed!

kwinkunks (approved):
Looks good after small changes. Can't see what might be causing the failed test on Python 3.10.

berland:
Probably a flaky thing that stopped 3.10. But there was a real pylint issue on 3.10 that needed attention, replaced  `distutils.dir_util.mkpath` with `pathlib.Path.mkdir`.

Issue #295, opened 2023-01-09
Version is hardcoded at 1.0


kwinkunks:
The `setup.py` file and `komodo/__init__.py` both hardcode the version as `1.0`.

We should move to semantic versioning using something like `setuptools-scm`.

kwinkunks:
Should deal with this as part of #296

kwinkunks:
In fact, `__init__.py` hardcodes a lot of package metadata, some of which is out of date or in direct conflict with the package (e.g. the license). Will deal with as part of #296.

kwinkunks:
Resolved by 2168d32b1ba73da8799d0b2f0eeeb4a3f81d9a9b

PR #298, opened 2023-01-09
modernize build system

### LINKED ISSUE #296: Modernize the build process
The build is not too complicated and should be quite easy to move to PEP621-style building with `pyproject.toml`. This would also allow us to gather all the build tool setup in one place, getting rid of `.isort.cfg`, for example.

kwinkunks:
Proof of concept for moving komodo to PEP 621 packaging, which uses only `pyproject.toml` (no `setup.py` or `setup.cfg`) and is meant to be easier to maintain. For example, all build tool options can go into the TOML file.

This changes how the project is built, requiring the `build` module. The wheel and sdist are produced by `python -m build`, but `pip install .` still works.

Note that the dynamic versioning depends on git tags and requires a `komodo/_version.py` file **which should not be kept under version control**.

Unit tests are passing but I don't know about CI yet. The TOML file will need to be updated when the pylint and show-version PRs are merged.

resolves #296 and #295


kwinkunks:
This PR resolves #296 and is ready for review.

komodo/cli.py
```diff
@@ -193,57 +193,156 @@ def cli_main():
...
+        required=True,
+        help="The name of the release, will be used as the name of the directory "
+        "containing the enable script and environment `root` directory.",
+    )
     required_args.add_argument(
         "--locations-config",
         type=str,
         required=True,
-        help="Path to locations.yml, a map of location to a server.",
+        help="Path to locations file, mapping locations to servers, " "in YAML format.",
```
berland, line 228: Merge strings.

komodo/cli.py
```diff
@@ -193,57 +193,156 @@ def cli_main():
...
+        type=str,
+        default="pip-cache",
+        help="The temporary directory used for downloads, e.g. by pip.",
+    )
+    optional_args.add_argument(
+        "--jobs",
+        "-j",
+        type=int,
+        default=1,
+        help="The number of parallel jobs to use for builds by cmake. Default: 1.",
```
berland, line 258: By using `parser = argparse.ArgumentParser(formatter_class=argparse.ArgumentDefaultsHelpFormatter) `the defaults are automatically injected into the help text.

komodo/cli.py
```diff
@@ -193,57 +193,156 @@ def cli_main():
...
-    optional_args.add_argument("--dry-run", "-n", action="store_true")
-
-    optional_args.add_argument("--cmake", type=str, default="cmake")
-    optional_args.add_argument("--pip", type=str, default="pip")
     optional_args.add_argument(
         "--virtualenv",
         type=str,
         default="virtualenv",
-        help="What virtualenv command to use",
+        help="The command to use for virtual environment constuction. "
```
berland, line 302: constuction -> construction

dev-requirements.txt
```diff
@@ -3,3 +3,4 @@ flake8
 isort
 pylint
 pytest
+build
```
berland, line 6: maybe sort the lines alphabetically

pyproject.toml
```diff
@@ -0,0 +1,98 @@
...
+]
+dependencies = [
+    "shell",
+    "PyYAML",
+    "ruamel.yaml",
+    "packaging",
+    "PyGithub >= 1.55",
+    "jinja2",
+    "requests",
+    "pysnyk",
```
berland, line 31: melikes having this sorted

komodo/cli.py
kwinkunks: That is very cool, I should have thought to look for it. Fixed this in the other PR (`cli-help-strings`)

komodo/cli.py
kwinkunks: :+1: Done in #317

pyproject.toml
kwinkunks: lol, done.

komodo/cli.py
kwinkunks: Good catch. Done in #308

kwinkunks:
@berland Can you see an issue with adding the `max-line-length` constraint to `pylint` *instead* of to `flake8`? Then we can dispense with the command-line option in the workflow and keep things neat.

berland:
> @berland Can you see an issue with adding the `max-line-length` constraint to `pylint` _instead_ of to `flake8`? Then we can dispense with the command-line option in the workflow and keep things neat.

Yes, default line length in flake8 is 79, so it would fail unless we set it.

.github/workflows/style.yml
```diff
@@ -35,12 +35,12 @@ jobs:
 
     - name: Run isort
       run: |
-        isort --check komodo tests *.py
+        isort --check komodo tests
 
     - name: Lint with flake8
       run: |
-        flake8
+        flake8 --max-line-length 88
```
xjules, line 42: Just a reminder for what was discussed: Shall we instead make it a pylint check for the max-line-length and source the value from `pyproject.toml`?

kwinkunks: See @berland's comment above: `flake8` needs the constraint otherwise it will fail.

komodo/__init__.py
```diff
@@ -1,12 +1,14 @@
...
+try:
+    VERSION = get_distribution(__name__).version
+except DistributionNotFound:
+    try:
+        from ._version import version as VERSION
+    except ImportError:
+        raise ImportError(
+            "Failed to find (autogenerated) _version.py. "
+            "This might be because you are installing from GitHub's tarballs, "
+            "use the PyPI ones."
```
berland, line 12: This comment might lead people to https://pypi.org/project/komodo/ which is not exacly not we want..

dev-requirements.txt
```diff
@@ -1,4 +1,5 @@
 black
+build
```
berland, line 2: What exactly in the other files refer to this package, I don't see how it gets engaged?

komodo/__init__.py
xjules: That's a good point @berland - missed that comment. Just to mention use `pip install . ` should suffice?

kwinkunks: Yes, good point; removed references to GitHub and PyPI... and trying to figure out something weird in my installation. Not exactly sure yet what someone will get when using `pip install .` on a cloned repo.

dev-requirements.txt
kwinkunks: In principle, `build` is required to build the project: https://packaging.python.org/en/latest/tutorials/packaging-projects/#generating-distribution-archives

But I see that `pip install .` seems to work either way. I generally build dists (wheel and sdist) with `python -m build` but I guess that's not strictly necessary if you're not uploading them anywhere.

Will do some more testing.

komodo/__init__.py
kwinkunks: In fact, `pip install .` works fine. But `komodo.__version__` won't do anything interesting until we start making release tags, if we ever do.

Anyway, leaving the message as "Failed to find autogenerated _version.py.".

dev-requirements.txt
kwinkunks: Have tested everything without `build` in my env and indeed it's all good, you just can't build distributions, which is fine. Removed it from the requirements.

xjules (approved):
Nice job @kwinkunks ! :rocket:

PR #299, opened 2023-01-11
Port m4 to jinja2

berland:


tests/test_enable_scipts.py
```diff
@@ -1,25 +1,25 @@
...
-                [
-                    "m4 {}".format(data.get("enable.m4")),
-                    "-D komodo_prefix={}".format(kmd_prefix),
-                    "-D komodo_pyver={}".format(kmd_pyver),
-                    "-D komodo_release={}".format(kmd_release),
-                    data.get(template),
-                ]
-            ).decode("utf-8")
-        )
+def _create_script(kmd_prefix, _kmd_pyver, kmd_release, target, template):
```
kwinkunks, line 9: Wondering if it would be 'safer' to have a function in `cli.py` (or elsewhere) that `cli.py` and this test both call? Because it seems like otherwise the test and the actual code could diverge?

berland: Yes, agreed!

kwinkunks (approved):
This looks v nice!

PR #300, opened 2023-01-12
Remove usage of renamer, use plain mv instead

### LINKED ISSUE #274: Refactor to use a different rename
When on Ubuntu 22.04, the `rename.ul` command is no longer available. See
https://unix.stackexchange.com/questions/716806/could-you-suggest-how-to-use-rename-ul-under-ubuntu-22-04-many-thanks

https://github.com/equinor/komodo/blob/1852315c1b80ff8b61996ad249c37a568adc4fc5/tests/test_cli.py#L57

berland:
Resolves #274

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -13,7 +13,7 @@ jobs:
       matrix:
         python-version: ['3.8', '3.10']
 
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-latest
 
     steps:
     - uses: actions/checkout@v3
@@ -52,7 +52,6 @@ jobs:
             --prefix /tmp/pfx \
             --release travis \
             --locations-config $(realpath ci/travis/locations.yml) \
-            --renamer rename.ul
 
     - name: Build libkmd.so for testing
       run: |
--- a/komodo/cli.py
+++ b/komodo/cli.py
@@ -117,18 +117,17 @@ def _main(args):
 
     print(f"Installing {args.release} to {args.prefix}")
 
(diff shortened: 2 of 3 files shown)


kwinkunks (commented):
I hadn't seen the `rename` command before so I'll leave approval to someone with more knowledge of any possible differences

Just wanted to **like** this change because the `rename` API is not the easiest to read and seemed like overkill here.

kwinkunks:
Just a nudge to @xjules @lars-petter-hauge and @hnformentin that I'm not comfortable approving this, but it seems good to me, and definitely improves the code readability.

hnformentin:
I am probably missing something: `ubuntu-latest` seems to run [ubuntu 22.04](https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners). So unpinning it, is actually no difference, correct?

You mentioned in #274 we may want to have`rename.ul` back when it reemerges. Could we have an issue to have `rename.ul` back pinned to here?

berland:
> I am probably missing something: `ubuntu-latest` seems to run [ubuntu 22.04](https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners). So unpinning it, is actually no difference, correct?

No, it is ubuntu 20.04 that is unpinned here.

> You mentioned in #274 we may want to have`rename.ul` back when it reemerges. Could we have an issue to have `rename.ul` back pinned to here?

Not needed, as after this PR we will not need `rename.ul` at all.

hnformentin (approved):
LGTM.

PR #302, opened 2023-01-13
Add json output option to komodo-non-deployed

### LINKED ISSUE #301: Add json output from komodo-non-deployed
For integration with Github action, it is relevant to be able to output in json format from `komodo-non-deployed`, in order to avoid the hackery in 
https://github.com/equinor/komodo-releases/pull/3043/files#diff-487281c14a5ff9f976920b3406c32dea2972482cce2375fa3d53ffe38adefcefR50

berland:
This eases integration with Github actions

Resolves #301

### DIFF
--- a/komodo/deployed.py
+++ b/komodo/deployed.py
@@ -1,5 +1,7 @@
 import argparse
+import json
 import os
+from typing import List, Optional
 
 from komodo.matrix import get_matrix_base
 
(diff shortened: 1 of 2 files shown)


komodo/deployed.py
```diff
@@ -62,6 +72,7 @@ def deployed_main():
         default=None,
         help="The maximum number of undeployed matrices to list.",
     )
+    parser.add_argument("--json", action="store_true", help="Get output in JSON format")
```
kwinkunks, line 75: Should perhaps add explicit default to CLI, and mention default in Help.

berland: The default is automatically set to `False`, but needed to add a `formatter_class` to get the default visible. Fixed.

kwinkunks (approved):
This looks good to me, assuming it provides the output you need in the GitHub workflow it services (I didn't check this)

berland:
> This looks good to me, assuming it provides the output you need in the GitHub workflow it services (I didn't check this)

Apparently me neither.. https://github.com/equinor/komodo/pull/311

PR #305, opened 2023-01-16
Add support for obtaining github tokens from env

berland:
Any jinja2 template found in the source part for packages in repository.yml will be expanded for environment variables.

### DIFF
--- a/komodo/fetch.py
+++ b/komodo/fetch.py
@@ -5,6 +5,8 @@ import argparse
 import os
 import sys
 
+import jinja2
+
 from komodo.package_version import (
     LATEST_PACKAGE_ALIAS,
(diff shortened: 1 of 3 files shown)


kwinkunks:
As mentioned, it occurred to me that some of the linting, or indeed the other scripts in komodo, might be checking formats etc, and I wonder if any of them should know about this extension to the syntax of our files?

My impression from a quick look is that there is currently no linting of the source happening in lint.py, but perhaps there should be.

If deemed non-urgent then this can be converted to a maintenance issue.

berland:
Linting seems to be fine with the environment variables.

kwinkunks (approved):
Nice. This looks like a solid update, continuing the modernization.

PR #306, opened 2023-01-16
Use Python 3.10 for style testing

berland:


kwinkunks:
Maybe take care of this at the same time https://github.com/equinor/komodo/issues/307

kwinkunks:
As mentioned, it occurred to me that some of the linting, or indeed the other scripts in `komodo`, might be checking formats etc, and I wonder if any of them should know about this extension to the syntax of our files?

My impression from a quick look is that there is currently no linting of the source happening in `lint.py`, but perhaps there should be. 

If deemed non-urgent then this can be converted to a maintenance issue.

kwinkunks (approved):
Makes sense, looks good.

PR #309, opened 2023-01-18
Support building for py310 and on rhel8

berland:


kwinkunks (commented):
These explicit checks should probably not be in the FOSS code. It feels weird to have business rules hardcoded into the linter --- especially such Equinor-centric rules.

Haven't given it tons of thought, but maybe best would be to have a config that specifies some default regexes which can be overridden. But this is probably overkill for now. Not sure what a 'light touch' solution that meets our needs would look like.

Not urgent so this can turn into an issue, by all means. I know there are several other things that need to be generalized like this.

kwinkunks:
Made #313 to deal with the de-Equinorization later.

komodo/lint.py
```diff
@@ -56,7 +56,7 @@ def lint_release_name(pkgfile):
             _kerr(
                 pkg=pkgfile,
                 err="Invalid release name suffix. "
-                "Must be of the most -pyXX or -pyXX-rhelY",
+                "Must be of the most -pyXX[X] or -pyXX-rhelY",
```
kwinkunks, line 59: I don't follow this wording. Does is mean "must be of the **form**" (I think it probably does?) or "must be **at** the most" (not sure this makes sense).

berland: Right, I did not look at the wording, but it does not make sense currently.. :)

berland: Changed "most" to "from"

kwinkunks (approved):
Looks good to me.

Issue #308, opened 2023-01-18
Add arguments for komodo-transpiler


xjules:
Currently `komodo-transpiler transpile` has hard-coded arguments for specifying python and rhel versions. We should be able to retrieve the versions from the cli arguments.
For example:
```sh
komodo-transpiler transpile
        --matrix-file=releases/matrices/${{ inputs.target }}.yml
        --python-ver=3.8
        --rhel-ver=7
        --output releases
```

kwinkunks:
Just a question because as mentioned in the stand-up, I find the language a bit confusing.

:question: Does the Release Manager want to maintain a single release **matrix** file, or multiple **release** files? I think it's the matrix file, is that correct?

I see that `komodo-transpiler` can go either way:

- The `combine` command takes multiple separate release files and creates a single matrix file.
- The `transpile` command takes a single matrix file and splits it into multiple release files.

kwinkunks:
I had a thought about how to add the urgently needed capability to produce a single 'cell' in a release matrix, without having to hardcode 'redhat' and 'python' into the tool in yet another place.

Maybe the interface could look like:

    komodo-transpiler transpile
        --matrix-file=releases/matrices/${{ inputs.target }}.yml
        --job py38,rhel7
        --output releases

This `--job` argument can then be split into the two 'coordinates' of the matrix. We are free to interpret them however, but I think the tool only needs to care that they are 'thing 1' and 'thing 2', right?

Maybe `release_transpiler.py` can also be refactored to not use variable names like `rhel_ver` and `py_ver`, but I didn't look at how hard that would be. And eventually we will want it to allow for an arbitrary number of 'coordinates' in `job` but we can deal with that when we need it.

What do you think @xjules and @hnformentin? Would this work?

hnformentin:
- In which cases will we only want to release one combination from the matrix? Will we have situations where we want two of them Helping to considerate @kwinkunks suggestion to use `--job` entry
- Which are possible places other than komodo where the hard-corded variables `RH-Versions` and `PY_VERSIONS` are used? Helping to answer if we can completely remove this hardcoded variables

@kwinkunks and @xjules do you know about this?

hnformentin:
Some changes to consider after discussion.
This are examples that should work:

(1) Base case:
komodo-transpiler transpile
        --matrix-file=release.yml
        --output releases
        rhel=7
        py=3.8

(2) Should be supported:
komodo-transpiler transpile
        --matrix-file=release.yml
        --output releases
[...]

PR #315, opened 2023-01-19
Add arguments to transpile

hnformentin:
Solves #308

### DIFF
--- a/komodo/matrix.py
+++ b/komodo/matrix.py
@@ -3,31 +3,33 @@ release matrix system. It should allow other parts of komodo to be capable of
 handling an arbitrary large and funky matrix, without having to guess and/or
 repeat itself."""
 import itertools
+import re
+from typing import Iterator, Sequence, Tuple
 
-RH_VERSIONS = ["7"]
-PY_VERSIONS = ["3.8"]
 
-
-def get_matrix():
-    """Return tuples of RHEL version and Python version, representing the
+def get_matrix(
+    rhel_versions: Sequence[str], py_versions: Sequence[str]
+) -> Iterator[Tuple[str, str]]:
+    """Return tuples of rhel version and Python version, representing the
     current release matrix."""
-    for product in itertools.product(RH_VERSIONS, PY_VERSIONS):
+    for product in itertools.product(rhel_versions, py_versions):
         rh_ver, py_ver = product
         yield (f"rhel{rh_ver}", f"py{py_ver.replace('.', '')}")
 
 
-def format_release(base, rhel_ver, py_ver):
+def format_release(base: str, rhel_ver: str, py_ver: str) -> str:
     """Format a base (e.g. a matrix file without the .yml suffix) such that it
     looks like a concrete release."""
     return f"{base}-{py_ver}-{rhel_ver}"
 
 
-def get_matrix_base(release):
+def get_matrix_base(release_name: str) -> str:
     """Return the base (e.g. matrix part of a concrete release). Should be the
-    inverse of format_release for actual, concrete matrix releases."""
-    for rhel_ver, py_ver in get_matrix():
-        suffix = format_release("", rhel_ver, py_ver)
-        if release.endswith(suffix):
-            return release.split(suffix, 1)[0]
+    inverse of format_release for actual, concrete matrix releases.
+    Hard-coded the suffix pattern '-py..-rhel.' or '-py...-rhel.'"""
+
+    suffix = format_release("", "rhel[0-9]", r"py\d{2,3}")
+    if re.search(suffix, release_name):
+        return re.split(suffix, release_name)[0]
     # no matrix suffix at all
-    return release
+    return release_name
--- a/komodo/release_transpiler.py
+++ b/komodo/release_transpiler.py
@@ -3,14 +3,28 @@
 import argparse
 import itertools
 import os
+import re
+from typing import Optional, Sequence
+
+import yaml
 
 from komodo.matrix import format_release, get_matrix
 from komodo.prettier import load_yaml, write_to_file
 
 
-def build_matrix_file(release_base, release_folder, builtins):
+def build_matrix_file(
+    release_base: str,
+    release_folder: str,
+    builtins: dict,
+    py_coords: Optional[Sequence[str]],
+) -> None:
+    """Combine release files from the release_folder into one single matrix_file."""
     files = {}
-    py_keys = [py_ver for _, py_ver in get_matrix()]
+    if py_coords is None:
+        py_keys = get_py_coords(release_base, release_folder)
+    else:
+        py_keys = [f"py{py_version.replace('.', '')}" for py_version in py_coords]
+
     for key in py_keys:
         files[key] = load_yaml(f"{release_folder}/{release_base}-{key}.yml")
 
@@ -32,9 +46,37 @@ def build_matrix_file(release_base, release_folder, builtins):
     write_to_file(compiled, f"{release_base}.yml", False)
 
 
-def _build(packages, py_ver, rhel_ver):
+def get_py_coords(release_base: str, release_folder: str) -> Sequence[str]:
+    """Get python versions of release files inside a given release_folder."""
+    filenames_with_prefix = sorted(
+        [
+            filename
+            for filename in os.listdir(release_folder)
+            if filename.startswith(release_base)
+        ]
+    )
+    len_release_base = len(release_base + "-")
+    irrelevant_suffix_length = len(".yml")
+    py_coords = [
+        filename[len_release_base:-irrelevant_suffix_length]
+        for filename in filenames_with_prefix
+    ]
+    return py_coords
+
+
+def _pick_package_versions_for_release(
+    packages: dict, rhel_ver: str, py_ver: str
+) -> dict:
+    """Consolidate the packages for a given combination of rhel and python version
+    into a dictionary."""
     release_dict = {}
-    for p, versions in packages.items():
+    for pkg_name, versions in packages.items():
+        try:
+            _check_version_exists_for_coordinates(versions, rhel_ver, py_ver)
+        except KeyError as err:
+            error_msg = f"{str(err)}. Failed for {pkg_name}."
+            raise KeyError(error_msg)
+
         if rhel_ver in versions:
             version = versions[rhel_ver][py_ver]
         elif py_ver in versions:
@@ -43,28 +85,85 @@ def _build(packages, py_ver, rhel_ver):
             version = versions
 
         if version:
-            release_dict[p] = version
+            release_dict[pkg_name] = version
     return release_dict
 
 
-def transpile_releases(matrix_file, output_folder):
+def _check_version_exists_for_coordinates(
+    pkg_versions: dict, rhel_coordinate: str, py_coordinate: str
+) -> None:
+    """Check the coordinates `rhel_ver` and `py_ver` input as arguments to
+    build a release against the release matrix file. Raise exceptions if
+    coordinates not found.
+    pkg_versions can take various levels as the examples show:
+        {
+            rhel7: # first_level
+                py36: 1.1.1, # second_level
+                py38: 2.1.1,
+            rhel8: # first_level
+                py36: 3.1.1, # second_level
+                py38: 4.1.1,
+        }
+        or:
+        {
+            py36: 1.1.1, # first level
+            py38: 2.1.1,
+        }
+        or:
+        {1.1.1}
+
+    """
+    first_level_versions = []
+    for version in pkg_versions:
+        first_level_versions.append(version)
+    if "rhel" in first_level_versions[0]:
+        # Both rhel and python versions can have different versions
+        if rhel_coordinate not in first_level_versions:
+            raise KeyError(f"Rhel version {rhel_coordinate} not found.")
+        second_level_versions = []
+        for version_py in pkg_versions[rhel_coordinate]:
+            second_level_versions.append(version_py)
+        if py_coordinate not in second_level_versions:
+            raise KeyError(
+                f"Python version {py_coordinate} not found for "
+                f"rhel version {rhel_coordinate}."
+            )
+    elif "py" in first_level_versions[0]:
+        # Only python has different versions
+        if py_coordinate not in first_level_versions:
+            raise KeyError(f"Python version {py_coordinate} not found.")
+
+
+def transpile_releases(matrix_file: str, output_folder: str, matrix: dict) -> None:
+    """Transpile a matrix file possibly containing different os and framework
+    versions (e.g. rhel6 and rhel7, py3.6 and py3.8).
+    Write one dimension file for each element in the matrix
+    (e.g. rhel7 and py3.8, rhel6 and py3.6)"""
+    rhel_versions = matrix["rhel"]
+    python_versions = matrix["py"]
+
     release_base = os.path.splitext(os.path.basename(matrix_file))[0]
     release_folder = os.path.dirname(matrix_file)
     release_matrix = load_yaml(f"{os.path.join(release_folder, release_base)}.yml")
-    for rhel_ver, py_ver in get_matrix():
-        release_dict = _build(release_matrix, py_ver, rhel_ver)
+    for rhel_ver, py_ver in get_matrix(rhel_versions, python_versions):
+        release_dict = _pick_package_versions_for_release(
+            release_matrix, rhel_ver, py_ver
+        )
         filename = f"{format_release(release_base, rhel_ver, py_ver)}.yml"
         write_to_file(release_dict, os.path.join(output_folder, filename))
 
 
 def combine(args):
     build_matrix_file(
-        args.release_base, args.release_folder, load_yaml(args.override_mapping)
+        args.release_base,
+        args.release_folder,
+        load_yaml(args.override_mapping),
+        args.py_coords,
     )
 
 
 def transpile(args):
-    transpile_releases(args.matrix_file, args.output_folder)
+    transpile_releases(args.matrix_file, args.output_folder, args.matrix_coordinates)
 
 
 def main():
@@ -112,9 +211,21 @@ Combine release files into a matrix file. Output format:
         required=True,
         help="File containing explicit matrix packages (default: None)",
     )
+    matrix_parser.add_argument(
+        "--py_coords",
+        help="""Comma delimitated list of python versions to be combined,
+        for example, "3.6,3.8" (without spaces).
+        If None, the release files in release-folder will be used to imply
+        the versions to combine.""",
+        type=lambda s: re.split(",", s),
+        required=False,
+        default=None,
+    )
 
     transpile_parser = subparsers.add_parser(
-        "transpile", description="Transpile a matrix file into separate release files."
+        "transpile",
+        description="Transpile a matrix file into separate release files.",
+        formatter_class=argparse.ArgumentDefaultsHelpFormatter,
     )
     transpile_parser.set_defaults(func=transpile)
     transpile_parser.add_argument(
@@ -126,6 +237,13 @@ Combine release files into a matrix file. Output format:
         "--output-folder",
         required=True,
         help="Folder to output new release files",
+    )
+    transpile_parser.add_argument(
+        "--matrix-coordinates",
+        help="Matrix to be transpiled, expected yaml format string.",
+        type=yaml.safe_load,
+        required=False,
+        default="{rhel: ['7'], py: ['3.8']}",
     )
     args = parser.parse_args()
     args.func(args)
--- a/tests/data/test_release_matrix.yml
+++ b/tests/data/test_release_matrix.yml
@@ -6,16 +6,23 @@ lib1:
   rhel7:
     py27: 0.1.2+builtin
     py36: 1.2.3+builtin
-    py38: 1.2.3+builtin
+    py38: 1.2.4+builtin
+    py310: 1.2.5+builtin
+  rhel8:
+    py38: 1.2.5+builtin
+    py310: 1.2.6+builtin
 lib2:
   py27: 1.2.3
   py36: 2.3.4
   py38: 2.3.4
+  py310: 2.3.6
 lib3:
   py27: 2.3.4
   py36: 3.4.5
   py38: 3.4.6
+  py310: 5.4.0
 lib4:
   py27: 3.4.5
   py36:
   py38:
+  py310: 4.0.0
--- a/tests/test_release_transpiler.py
+++ b/tests/test_release_transpiler.py
@@ -1,8 +1,13 @@
 import os
 
+import pytest
 import yaml
 
-from komodo.release_transpiler import build_matrix_file, transpile_releases
+from komodo.release_transpiler import (
+    build_matrix_file,
+    get_py_coords,
+    transpile_releases,
+)
 from tests import _get_test_root
 
 builtins = {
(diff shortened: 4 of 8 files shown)


kwinkunks:
I had a thought about how to add the urgently needed capability to produce a single 'cell' in a release matrix, without having to hardcode 'redhat' and 'python' into the tool in yet another place.

Maybe the interface could look like:

    komodo-transpiler transpile
        --matrix-file=releases/matrices/${{ inputs.target }}.yml
        --job py38,rhel7
        --output releases

This `--job` argument can then be split into the two 'coordinates' of the matrix. We are free to interpret them however, but I think the tool only needs to care that they are 'thing 1' and 'thing 2', right?

Maybe `release_transpiler.py` can also be refactored to not use variable names like `rhel_ver` and `py_ver`, but I didn't look at how hard that would be. And eventually we will want it to allow for an arbitrary number of 'coordinates' in `job` but we can deal with that when we need it.

What do you think @xjules and @hnformentin? Would this work?

kwinkunks:
:point_up: Maybe I should move this comment to the issue??

komodo/matrix.py
```diff
@@ -24,10 +22,12 @@ def format_release(base, rhel_ver, py_ver):
 
 def get_matrix_base(release):
     """Return the base (e.g. matrix part of a concrete release). Should be the
-    inverse of format_release for actual, concrete matrix releases."""
-    for rhel_ver, py_ver in get_matrix():
-        suffix = format_release("", rhel_ver, py_ver)
-        if release.endswith(suffix):
-            return release.split(suffix, 1)[0]
+    inverse of format_release for actual, concrete matrix releases.
+    Hard-coded the suffix pattern '-py..-rhel.' or '-py...-rhel.'"""
```
kwinkunks, line 26: I wonder if we really need to hardcode details of our particular matrix into here? For example, can we define a more general pattern that ~all release names will adhere to?

Maybe releases will all look like `1111.11.11a11-pppp11-os11-other111`? At least that would avoid mentioning `py` and `rhel`. (Or maybe there's some other pattern, this is just an example.) The beauty of regex is that we can avoid having to be too specific. At its simplest, maybe it's just "everything before the first hyphen".

Alternatively, maybe the details of the matrix can go in a site config file? Or get passed in to the CLI that's making this query?

komodo/matrix.py
```diff
@@ -24,10 +22,12 @@ def format_release(base, rhel_ver, py_ver):
...
+
+    if re.search("-py..-rhel.", release):
```
kwinkunks, line 28: You should be able to use capturing groups (using parentheses) and more specific wildcards, like `[0-9]` or `\d` for digits, to capture everything 'at once' here. Something like `\d{2,3}` means "2 or 3 digits", for example. 

Here's one of those sites you can practise regex on: https://regex101.com/ -- and we can take a look together if you want.

komodo/matrix.py
hnformentin: Thanks for the comment. 
The format of a release is governed by [format_release](https://github.com/equinor/komodo/blob/f20735c60b84620ee3d08b15afd8f05331e30ddb/komodo/matrix.py#L19) function. I changed the code to keep the link with that function, which is a little better than pure hardcoded.
Nevertheless, I cannot see how we could get rid of hardcoding completely.

There is [one test](https://github.com/equinor/komodo/blob/f8fc14e46b0cfbc958d0ccb17a190dcbadd55fab/tests/test_matrix.py#L17) that would not pass with the `-` rule.

On the alternative you mention, do you suggest to have another argument to `transpile`? For example, we would have 4 arguments (or 2) for the current coordinates:  separate `os` from `os_version` and `prog_language` from `prog_language_version`?

komodo/matrix.py
hnformentin: Excellent tip!

hnformentin:
I added some commits with improvements from the review version: change order, remove repetition, fix black. 
Will rebase all in one commit before merging.

komodo/release_transpiler.py
```diff
@@ -4,13 +4,15 @@
...
+
 from komodo.matrix import format_release, get_matrix
 from komodo.prettier import load_yaml, write_to_file
 
 
-def build_matrix_file(release_base, release_folder, builtins):
+def build_matrix_file(release_base, release_folder, builtins, coords):
     files = {}
-    py_keys = [py_ver for _, py_ver in get_matrix()]
+    py_keys = [f"py{py_version.replace('.', '')}" for py_version in coords["py"]]
```
kwinkunks, line 15: My thought was that by using the `coords` dict, I think it should be possible to get rid of `py_ver` and `rhel_ver` completely. Instread of passing those variable around, pass the `coords` dict around. Maybe I'm missing something and it's not possible, but it seems like the `coords` dict should be able to have entries like `macos_ver:['10.15', '10.16']` and the code will just accommodate it as another constraint to add to the release string. This would mean looping over both the keys and the values to generate all the various release name strings.

(I _think_ this would mean we could also do things like this with another entry in the `coords` dict > https://github.com/equinor/komodo/pull/331 )

kwinkunks (changes requested):

komodo/release_transpiler.py
```diff
@@ -43,31 +46,65 @@ def _build(packages, py_ver, rhel_ver):
...
     )
 
 
 def transpile(args):
-    transpile_releases(args.matrix_file, args.output_folder)
+    transpile_releases(args.matrix_file, args.output_folder, args.coords)
 
 
 def main():
+    default_coords = "{rhel: ['7'], py: ['3.8']}"
```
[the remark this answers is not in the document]
hnformentin: If I understand you correctly, we actually had this default in `matrix.py` as 
```
RH_VERSIONS = ["7"]
PY_VERSIONS = ["3.8"]
```
and this PR is substituting it to a single place, as you suggested in the last part of the comment.
Did I misunderstand something?

komodo/release_transpiler.py
hnformentin: I fully understand your point.
In order to keep the scope of the issue, I would prefer to use another issue i.e. #332 to consider turning komodo agnostic to os and dev system.

kwinkunks:
I think it's worth adding the type of the return to the functions as well. You should also add docstrings -- now that you've taken the time to figure out what these things do, I think it will help the next person. (Is the type of the YAML dict really `OrderedDict`? I would have thought `dict` would do, since the basic type is ordered since Py 3.6. Maybe just check the type or see what MyPy says.)

I'm still bothered by the very specific regex in `matrix.py`, which looks for Python and RHEL. Is it really not possible to generalize the pattern? Or, perhaps better, pass in the matrix itself (`coords` I guess, see below) and figure out the 'base' from there? And again, you can use parentheses to capture the group you want and avoid needing `re.split()`. 

I've a feeling `coords` should be called `matrix`, what do you think?

hnformentin:
> I think it's worth adding the type of the return to the functions as well. You should also add docstrings -- now that you've taken the time to figure out what these things do, I think it will help the next person. 

Nice, I tried to add docstring and types of returns. Hope it helps the next to come across the code.

> (Is the type of the YAML dict really `OrderedDict`? I would have thought `dict` would do, since the basic type is ordered since Py 3.6. Maybe just check the type or see what MyPy says.)

I checked the type within a test and it is indeed OrderedDict. Also tried mypy (which turned up to be useful to identify an non-imported package), and it seems to be fine with it.

> I'm still bothered by the very specific regex in `matrix.py`, which looks for Python and RHEL. Is it really not possible to generalize the pattern? Or, perhaps better, pass in the matrix itself (`coords` I guess, see below) and figure out the 'base' from there? And again, you can use parentheses to capture the group you want and avoid needing `re.split()`.

For sure there is way to remove and generalise the pattern, but as it is not the focus of this issue, I will let this work to the issue #332 . I refer the issue so the one picking it can access the comments/suggestions here.

> I've a feeling `coords` should be called `matrix`, what do you think?

I think `coords` is slightly better because calling `matrix` would potentially generate confusion if it is the matrix file, for example. So I would stick with `coords`. Maybe `axis` would be another possibility, or `row` and `column`, though this last option could be even more confusing!

hnformentin:
Aiming to remove the dependence of `combine` of the hard-coded PY_VERSION variable that we had in `matrix.py`, I ended up choosing to specify the argument `--py_coord` -  a list of python versions to be considered. This is an optional arg, and if it is not defined, we will search which are the files in the release_folder and combine all of those that match the release_base.

komodo/release_transpiler.py
```diff
@@ -112,6 +180,15 @@ def main():
         required=True,
         help="File containing explicit matrix packages (default: None)",
     )
+    matrix_parser.add_argument(
```
kwinkunks, line 183: I thought the Python version (and any arbitrary matrix 'coordinates') could be encoded in the `coords` argument?

komodo/release_transpiler.py
```diff
@@ -43,28 +71,68 @@ def _build(packages, py_ver, rhel_ver):
...
 
         if version:
-            release_dict[p] = version
+            release_dict[pkg_name] = version
     return release_dict
 
 
-def transpile_releases(matrix_file, output_folder):
+def _check_consistency_in_versions(
+    versions: OrderedDict, pkg_name: str, rhel_ver: str, py_ver: str
```
kwinkunks, line 79: Does this definitely need to be of type `OrderedDict`? It seems like `dict` should suffice. I think the object is coming from `load_yaml`, which I think uses `ruamel`, but maybe just check its type, and/or see if you can 'get away' with just a dict (which will be ordered anyway).

komodo/release_transpiler.py
```diff
@@ -32,9 +44,25 @@ def build_matrix_file(release_base, release_folder, builtins):
...
+            if filename.startswith(release_base)
+        ]
+    )
+    len_release_base = len(release_base + "-")
+    py_coords = [filename[len_release_base:-4] for filename in filenames_with_prefix]
+    return py_coords
+
+
+def _build(packages: OrderedDict, rhel_ver: str, py_ver: str) -> dict:
+    """Consolidate the packages for a given python version into a dictionary."""
```
kwinkunks, line 62: The function also takes an `rhel` version, so maybe the docstring should reflect that.

Did you try just passing in arbitrary coordinates here? I still think it should be possible to make this function agnostic to the matrix. (E.g. have it consume `coords` instead of `rhel_ver` and `py_ver` and loop over its keys.)

komodo/release_transpiler.py
hnformentin: As the code of the function `combine` is of now, it needs only the python versions.
I guess this is due to historical reasons.
The options were:
1. keep `PY_VERSION=[3.8]` hardcoded as it was previously;
2. pass `coords` even though `rhel_ver` is not used, which means we are passing an unnecessary argument;
3. let only `--py-coords` as an optional argument;
4. remove the function `combine` as a whole - it is not used by us now. I preferred not because it maybe an useful feature for users outside scout or maybe can be useful in the future.

I preferred the third option, but would you prefer the second one?

komodo/release_transpiler.py
hnformentin: Yes, indeed `OrderedDict` could be substituted overall. Thanks!

hnformentin:
I cannot see the specific changed you required, from `Merging can be performed automatically once the requested changes are addressed`. Can you please point out them into a comment or help me to find them otherwise?

tests/test_release_transpiler.py
```diff
@@ -32,13 +37,38 @@ def test_build_release_matrix(tmpdir):
         assert release_matrix["lib2"] == "2.3.4"
 
 
-def test_transpile(tmpdir):
+@pytest.mark.parametrize("coords", [({"py": ["3.8"], "rhel": ["7"]})])
```
kwinkunks, line 40: Do the tests produce the correct results if you use (say) `{"py": ["3.8", "3.10"], "rhel": ["7", "8"]}`?

komodo/release_transpiler.py
kwinkunks: I think we should avoid adding Equinor-specific arguments so I think it should accept the same matrix definition (i.e. `coords` --- which should maybe be called `matrix`).

If we don't want to refactor it now because we don't use it, maybe we should deprecate it.

komodo/matrix.py
kwinkunks: Eventually I think `format_release()` could look more like this:

```python
def format_release(base, matrix):
    for coords in product(*matrix.values()):
        coord_strings = [f"{name}{coord}" for name, coord in zip(matrix.keys(), coords)]
        yield '-'.join([base] + coord_strings)
```

Then be called like...

```python
matrix = {"py": ["3.8", "3.10", "3.11"], "rhel": ["7"], "loc": ["Azure", "Onprem"]}

for fname in format_release('baseybase', matrix):
    print(fname)
```

This produces:

```
baseybase-py3.8-rhel7-locAzure
baseybase-py3.8-rhel7-locOnprem
baseybase-py3.10-rhel7-locAzure
baseybase-py3.10-rhel7-locOnprem
baseybase-py3.11-rhel7-locAzure
baseybase-py3.11-rhel7-locOnprem
```

This way, it feels like we have lots more flexibility because the code never needs to know what our matrix names and coordinates are --- we could add Ruby or switch to Ubuntu without changing anything.

If this is all too much to ask of this PR, no worries --- I just wanted to clarify what I had in my mind! (For me as much as anything :)

tests/test_release_transpiler.py
hnformentin: yeap, I included the suggestion in the test and updated the test data `test_release_matrix.yml` for that.

komodo/release_transpiler.py
kwinkunks: Okay, if what you have does what's needed then :+1:

komodo/release_transpiler.py
kwinkunks: Is it possible to avoid adding another specific 'coordinate' argument here?

komodo/release_transpiler.py
kwinkunks: Right -- I think this point is out of date wrt the code, but either way, I will just focus on #332 for this sort of thing.

komodo/release_transpiler.py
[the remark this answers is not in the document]
kwinkunks: I think this is outdated; will double-check in latest.

komodo/matrix.py
kwinkunks: If this won't change, then I'll let you hit *Resolve conversation* to close this thread.

kwinkunks:
> I cannot see the specific changed you required, from `Merging can be performed automatically once the requested changes are addressed`. Can you please point out them into a comment or help me to find them otherwise?

Not totally sure but I _think_ it's referring to the **Resolve conversation** items in the review. I think it's okay for you to close those on things as they are resolved (or rejected!).

komodo/release_transpiler.py
hnformentin: > Is it possible to avoid adding another specific 'coordinate' argument here?

I understand you refer to 'coordinate' to `--py_coords` argument. If so, it is an argument keep the previous functionaty of `combine`. But it is optional and by default the versions to be combined will be retrieved in the `release_folder`.

As it is an optional argument, I don't see the problem of the solution. Do you have something particularly in mind?
Remove the option and let only the default behaviour?

komodo/matrix.py
hnformentin: That is indeed a good point. I renamed the entry argument as `matrix` and referred your specific suggestion in #332 . Will certainly help who picks that issue.

komodo/release_transpiler.py
kwinkunks: If there's really no way to avoid adding an argument to the interface that we will have to deprecate when we eventually fix all this Py/RHEL stuff, then it has to be done. I trust that you've considered all the alternatives.

hnformentin:
For reference, the feature to make komodo more agnostic and let the argument `matrix` to be more flexible was discussed a couple of times within the squad. As it is referred in some comments in this PR, we decided to make another issue to enable this future proof flexibility #332 as this was not the priority at the moment. Thanks for the comments in the review so far @kwinkunks 🤓

komodo/release_transpiler.py
```diff
@@ -32,9 +44,26 @@ def build_matrix_file(release_base, release_folder, builtins):
...
+            for filename in os.listdir(release_folder)
+            if filename.startswith(release_base)
+        ]
+    )
+    len_release_base = len(release_base + "-")
+    py_coords = [filename[len_release_base:-4] for filename in filenames_with_prefix]
+    return py_coords
+
+
+def _build(packages: dict, rhel_ver: str, py_ver: str) -> dict:
```
valentin-krasontovitsch, line 61: can we rename this function? the idea is that we based on rh and py versions pick a specific version for each package, right? what do you think about:

```suggestion
def _package_version_picker(packages: dict, rhel_ver: str, py_ver: str) -> dict:
```

komodo/release_transpiler.py
```diff
@@ -43,28 +72,68 @@ def _build(packages, py_ver, rhel_ver):
             version = versions
 
         if version:
-            release_dict[p] = version
+            release_dict[pkg_name] = version
     return release_dict
 
 
-def transpile_releases(matrix_file, output_folder):
+def _check_consistency_in_versions(
```
valentin-krasontovitsch, line 79: yet another renaming suggestion. the versions in the function name seem to point to the versions argument, but we're not checking that those are consistent - rather we're checking that the given "coordinates" (rhel version + py version) are present in the package versions... thankfully this is so simple XD
how about something like

```suggestion
def _check_version_exists_for_coordinates(
```

and perhaps replace `rhel_ver` with `rhel_coordinate`

but this is maybe just bike-shedding, i dunno, the naming in this context is difficult, as i feel like we don't really have conceptual names down for what we're working with...

komodo/release_transpiler.py
```diff
@@ -43,28 +72,68 @@ def _build(packages, py_ver, rhel_ver):
...
+    versions: dict, pkg_name: str, rhel_ver: str, py_ver: str
+) -> None:
+    """Check the coordinates `rhel_ver` and `py_ver` inputed as arguments to
+    build a release against the release matrix file. Raise exceptions if
+    coordinates not found.
+    """
+    first_level_versions = []
+    for version in versions:
+        first_level_versions.append(version)
```
valentin-krasontovitsch, line 88: the name `first_level_versions` points to something magical (what is the first level)? I see also that we don't assume which comes first. that's i think too liberal - i think we can pin down the sequence of coordinates, i think that is consistent with the whole "matrix" analogy. so perhaps rename to `rhel_versions`, or if you choose to rename `rhel_ver` to `rhel_coordinate`, then this could be renamed to `rhel_coordinate_values` - or if you wanna move to a more general thing, the arguments could be called `first_coord`, `second_coordinate`, and correspondingly `first_level_versions` could become `first_coordinate_values`.

valentin-krasontovitsch (changes requested):
aaahhhh sooo many comments, i'm sorry. some of it, the typing stuff, is formalities. some of it is probably bike-shedding / arguing about naming. other things might be more substantial...

tests/data/test_release_matrix.yml
```diff
@@ -6,16 +6,23 @@ lib1:
...
 lib3:
   py27: 2.3.4
   py36: 3.4.5
   py38: 3.4.6
+  py310: 5.4.0
 lib4:
   py27: 3.4.5
   py36:
   py38:
+  py310:
```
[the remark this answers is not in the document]
hnformentin: Interesting, if I only add a new line, I get the error:

> <stdin>:12: trailing whitespace.
>   warning: 1 line adds whitespace errors.

So I need to add a version for python 3.10 example.

Do you know how I should do to let it empty? Maybe only add `py310: ""` was required?

tests/test_release_transpiler.py
```diff
@@ -21,7 +26,7 @@ def test_build_release_matrix(tmpdir):
     release_base = "2020.01.a1"
     release_folder = os.path.join(_get_test_root(), "data/test_releases/")
     with tmpdir.as_cwd():
-        build_matrix_file(release_base, release_folder, builtins)
+        build_matrix_file(release_base, release_folder, builtins, ["3.8"])
```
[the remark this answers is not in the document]
hnformentin: I think it is worth trying to write the test, at least it will add clarity to the functionality.
I will give it a try.

tests/test_release_transpiler.py
```diff
@@ -32,13 +37,41 @@ def test_build_release_matrix(tmpdir):
...
-                assert os.path.isfile(
-                    "{}-{}-{}.yml".format(release_base, py_ver, rhel_ver)
-                )
+        transpile_releases(release_file, os.getcwd(), matrix)
+        rhel_ver_str = f"rhel{matrix['rhel'][0]}"
+        py_ver_str = f"py{matrix['py'][0].replace('.', '')}"
+
+        assert os.path.isfile(
+            "{}-{}-{}.yml".format(release_base, py_ver_str, rhel_ver_str)
+        )
```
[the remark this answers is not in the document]
hnformentin: Good point!

tests/test_release_transpiler.py
```diff
@@ -32,13 +37,41 @@ def test_build_release_matrix(tmpdir):
...
+@pytest.mark.parametrize(
+    "matrix",
+    [
+        ({"py": ["3.8"], "rhel": ["7"]}),
+        ({"py": ["3.7"], "rhel": ["7"]}),
+        ({"py": ["3.6"], "rhel": ["5"]}),
+    ],
+)
+def test_validate_versions_in_args_with_release_file(matrix):
+    release_file = os.path.join(_get_test_root(), "data", "test_release_matrix.yml")
```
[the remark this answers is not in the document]
hnformentin: nice, I need to think about it.

komodo/release_transpiler.py
```diff
@@ -32,9 +44,26 @@ def build_matrix_file(release_base, release_folder, builtins):
...
+    """Get python versions of release files inside a given release_folder."""
+    filenames_with_prefix = sorted(
+        [
+            filename
+            for filename in os.listdir(release_folder)
+            if filename.startswith(release_base)
+        ]
+    )
+    len_release_base = len(release_base + "-")
+    py_coords = [filename[len_release_base:-4] for filename in filenames_with_prefix]
```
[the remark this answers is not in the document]
valentin-krasontovitsch: perfect, thank you!

komodo/release_transpiler.py
```diff
@@ -127,6 +205,13 @@ def main():
         required=True,
         help="Folder to output new release files",
     )
+    transpile_parser.add_argument(
+        "--matrix",
+        help="Matrix to be transpiled.",
```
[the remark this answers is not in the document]
valentin-krasontovitsch: > And to be honest I think it would be ok once we specify matrix-coordinates.

sure, sure - we can skip renaming the first one. don't wanna upset all the users of komodo out there : )

> But I did not play with it

i tried it out, sadly it doesn't display the default value - if the arg parse lib doesn't support it, you could of couse just pull out the default value into a variable...

tests/data/test_release_matrix.yml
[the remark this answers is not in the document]
valentin-krasontovitsch: hmm i think it's an editor thingy, this missing new line - i would hope that you don't need to specify an actual version, or an empty string or empty object (`{}` in yaml) for the last python version, while leaving the other python versions empty. who is it that complains?
sadly i never tried to understand the technical details around this "newline missing" complaint, so i don't know how to make it work exactly, but we can try to look at it together

komodo/release_transpiler.py
valentin-krasontovitsch: > Do you agree with this one?

sure, sounds good!

ohhh i haven't worked with this feature yet from the coder point of view, so i'm not sure, probably you will have to do the rename in other places yourself.

komodo/release_transpiler.py
valentin-krasontovitsch: after looking at the input we allow here, it becomes clear that more meaning cannot be added to the names, as we are way too permissive with the inputs we allow (I think). will create a separate issue to look at the input validation.

hnformentin: We chatted that the better approach would be to include a longer comment, explaining that `pkg_versions` can take various formats.

hnformentin (commented):
Some changes and corresponding replies so far!

tests/test_release_transpiler.py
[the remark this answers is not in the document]
valentin-krasontovitsch: looks good, thanks!

komodo/release_transpiler.py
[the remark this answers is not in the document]
valentin-krasontovitsch: nice find! i'm still missing that it expects yaml formatted input in the help text. although i don't know how much it will help, and the default provides a working example, and makes it obvious how to extend it.

oh one point though that i forgot to mention before, you can remove the quotes around `rhel` and `python` in the default value, according to my testing at least

tests/test_release_transpiler.py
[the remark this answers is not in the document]
valentin-krasontovitsch: i don't see how you refactored the test, maybe it's lost in between different diffs, but i cannot see whether calling `build_matrix_file` without passsing the last argument is tested - have you covered that code path somewhere?

komodo/release_transpiler.py
```diff
@@ -43,28 +72,68 @@ def _build(packages, py_ver, rhel_ver):
...
 
         if version:
-            release_dict[p] = version
+            release_dict[pkg_name] = version
     return release_dict
 
 
-def transpile_releases(matrix_file, output_folder):
+def _check_consistency_in_versions(
+    versions: dict, pkg_name: str, rhel_ver: str, py_ver: str
```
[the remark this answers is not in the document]
hnformentin: I changed the code to use the what you suggested, but actually did not understand how strings from errors exception work. I will add a comment on that on the new code.

tests/test_release_transpiler.py
```diff
@@ -32,13 +37,83 @@ def test_build_release_matrix(tmpdir):
...
+        ),
+    ],
+    ids=["Pass for all packages", "Fail", "Fail"],
+)
+def test_check_version_exists_for_coordinates(matrix, reason_for_failure):
+    release_file = os.path.join(_get_test_root(), "data", "test_release_matrix.yml")
+    try:
+        transpile_releases(release_file, os.getcwd(), matrix)
+    except KeyError as exception_info:
+        assert reason_for_failure in str(exception_info)
```
hnformentin, line 112: If I stop the test at this point and `print(str(exception_info))`, I get:
`"'Rhel version rhel5 not found for package lib1.'"`
I don't understand why it has `'"` and how to remove `'`.
Do you have an idea for that?

valentin-krasontovitsch: oh ehm hmmm not sure - it's the quotation marks around the whole thing that you want to get rid of? how do you get the output / print ? do you use a debugger? is the test failing because of the quotation marks?

in general, i think you could match for a smaller part of the error message, to make the test less brittle when somebody should decide to change the error message - one could argue that we don't really care about the wording, but rather that some key information is present, like

```python
except KeyError as exception_info:
    for info in expected_info:
        assert info in str(exception_info)
```

where `expected_info` would for instance be `["py37", "rhel7","lib1"]` for the second test (and you could still keep the full text reason as the test id, for better understanding of what the test is testing / why it should fail)

hnformentin: Nice comment, I use the pattern all(), but have same effect.

valentin-krasontovitsch: great!

komodo/release_transpiler.py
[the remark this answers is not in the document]
valentin-krasontovitsch: perfect, thank you!

valentin-krasontovitsch (approved):

hnformentin:
This changes small changes after approval were made after discussion with the reviewer 🚀

PR #317, opened 2023-01-20
add help strings for all CLIs

### LINKED ISSUE #288: Most of the CLI options do not have a help string or type
The inline help is missing for most of the 20 arguments, including 4 of the 5 required arguments, in the main CLI (`kmd`, defined in `cli.py`). Only `locations-config`, `virtualenv` and `workspace` have a `help` string, and at least 5 of them don't have a `type` or `default`.

kwinkunks:
Most of the CLI tools in Komodo had help strings, but the main one, `kmd`, was missing most of them. This commit makes them all more consistent with each other, and fills in some gaps.

resolves #288

[edit: preformatted text omitted by git CLI]

### DIFF
--- a/komodo/cli.py
+++ b/komodo/cli.py
@@ -190,57 +190,152 @@ def cli_main():
         "Automatically, reproducibly, and testably create software "
         "distributions.",
         add_help=False,
+        formatter_class=argparse.ArgumentDefaultsHelpFormatter,
     )
 
-    parser.add_argument("pkgs", type=YamlFile())
-    parser.add_argument("repo", type=YamlFile())
+    parser.add_argument(
+        "pkgs",
+        type=YamlFile(),
+        help="A Komodo release file mapping package name to version, "
+        "in YAML format.",
+    )
+    parser.add_argument(
+        "repo",
+        type=YamlFile(),
+        help="A Komodo repository file, in YAML format.",
+    )
 
     required_args = parser.add_argument_group("required named arguments")
 
-    required_args.add_argument("--prefix", "-p", type=str, required=True)
-    required_args.add_argument("--release", "-r", type=str, required=True)
+    required_args.add_argument(
+        "--prefix",
+        "-p",
+        type=str,
+        required=True,
+        help="The path of the directory in which you would like to place the "
+        "built environment.",
+    )
+    required_args.add_argument(
+        "--release",
+        "-r",
+        type=str,
+        required=True,
+        help="The name of the release, will be used as the name of the directory "
+        "containing the enable script and environment `root` directory.",
+    )
     required_args.add_argument(
         "--locations-config",
         type=str,
         required=True,
-        help="Path to locations.yml, a map of location to a server.",
+        help="Path to locations file, mapping locations to servers, in YAML format.",
     )
 
     optional_args = parser.add_argument_group("optional arguments")
 
     optional_args.add_argument(
-        "--help", "-h", action="help", help="show this help message and exit"
+        "--help",
+        "-h",
+        action="help",
+        help="Show this help message and exit.",
     )
-    optional_args.add_argument("--tmp", "-t", type=str)
-    optional_args.add_argument("--cache", "-c", type=str, default="pip-cache")
-    optional_args.add_argument("--jobs", "-j", type=int, default=1)
-
-    optional_args.add_argument("--download", "-d", action="store_true")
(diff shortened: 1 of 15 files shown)


kwinkunks:
Some review comments from @berland on #298 were added to this PR because they were more relevant.

komodo/release_transpiler.py
```diff
@@ -93,30 +94,38 @@ def main():
...
     py38 : 5.11.13+builtin""",
-        formatter_class=RawTextHelpFormatter,
+        formatter_class=argparse.RawTextHelpFormatter,
     )
     matrix_parser.set_defaults(func=combine)
     matrix_parser.add_argument(
-        "--release-base", help="Name of the release to handle", required=True
+        "--release-base",
+        required=True,
+        help="Name of the release to handle (default: None)",
```
hnformentin, line 103: Why "default: None", but not explicit in `default`? 
combine -> build_matrix_file seems to expect a name for the path.

Same comment for the 3 instances below also.

komodo/show_version.py
```diff
@@ -33,8 +33,9 @@ def get_release() -> str:
             """
             No active komodo environment found.
 
-            Either run this command from an active komodo or komodoenv environment,
-            or provide an environment manifest file path with the --path option.\
+            Either run this command from an active komodo or komodoenv
+            environment, or provide an environment manifest file path with the
+            --manifest-file option.\
```
hnformentin, line 38: I did not follow up this change. Why --path -> --manifest-file?

komodo/snyk_reporting.py
```diff
@@ -41,14 +41,36 @@ def main() -> None:
 
 def parse_args(args: Dict[str, str]) -> argparse.Namespace:
     parser = argparse.ArgumentParser(
-        description="Test a release for security and license issues"
+        description="Test a release for security and license issues",
+        formatter_class=argparse.ArgumentDefaultsHelpFormatter,
+    )
+    parser.add_argument(
+        "--orgid", type=str, required=True, help="The Snyk organization ID"
```
hnformentin, line 48: For readability, one key with its line?

komodo/snyk_reporting.py
```diff
@@ -41,14 +41,36 @@ def main() -> None:
...
+    )
+    parser.add_argument(
+        "--repo",
+        type=YamlFile(),
+        required=True,
+        help="A Komodo repository file, in YAML format.",
```
hnformentin, line 54: small k in Komodo. Same comment for the instance below.

hnformentin (commented):
Good initiative to improve documentation.

komodo/cli.py
```diff
@@ -191,59 +191,157 @@ def cli_main():
...
-    parser.add_argument("pkgs", type=YamlFile())
-    parser.add_argument("repo", type=YamlFile())
+    parser.add_argument(
+        "pkgs",
+        type=YamlFile(),
+        help="A Komodo release file mapping package name to version, "
+        "in YAML format.",
+    )
+    parser.add_argument(
+        "repo", type=YamlFile(), help="A Komodo repository file, in YAML format."
```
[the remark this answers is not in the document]
kwinkunks: good catch

komodo/release_transpiler.py
kwinkunks: Well spotted! This parser uses `argparse.RawTextHelpFormatter` to handle showing the expected output file format properly. There seems to be no other way to achieve that (newlines are removed), and I couldn't fine a way to combine it with `argparse.ArgumentDefaultsHelpFormatter`, which normally handles the transclusion of the degfaults. So we end up with this -- not great, but produces what we need.

komodo/show_version.py
kwinkunks: Håvard felt that `manifest-file` (or similar) was more explicit than `path`, and I agree, so I changed the CLI argument... but forgot to change this mention of it.

komodo/snyk_reporting.py
kwinkunks: yes, missed that one :+1:

komodo/snyk_reporting.py
kwinkunks: I think most mentions are now uppercase so I think I'll leave this for now. I don't think there's any ambiguity.

berland (approved):
👍🏻

PR #318, opened 2023-01-20
remove locations-config from required args

### LINKED ISSUE #314: Make locations-config an optional argument for kmd
Currently the `kmd` CLI requires that the user creates a `locations.yml` file (it can have any name) with contents like:

```
locations:
  foo: my-computer.foo.example.com
```

This sets the value of `ERT_LSF_SERVER` in the scripts `local` and `local.csh`, which has content like:

```shell
host=$(hostname)
location=$(echo $host | cut -d "-" -f 1)

case ${location} in
   foo)
      export ERT_LSF_SERVER=my-computer.foo.example.com
        ;;
esac
```

The tool **requires** this file, even though it is probably not useful for most other users of the tool (hypothetical users at the moment, but...)

If we do change this, the README will need changing too.

kwinkunks:
resolves #314

### DIFF
--- a/komodo/cli.py
+++ b/komodo/cli.py
@@ -88,13 +88,16 @@ def _main(args):
 
     create_enable_scripts(komodo_prefix=tmp_prefix, komodo_release=args.release)
 
-    with open(args.locations_config, mode="r", encoding="utf-8") as defs, open(
-        Path(args.release) / "local", mode="w", encoding="utf-8"
-    ) as local_activator, open(
-        Path(args.release) / "local.csh", mode="w", encoding="utf-8"
-    ) as local_csh_activator:
-        defs = yml.safe_load(defs)
-        local.write_local_activators(data, defs, local_activator, local_csh_activator)
+    if args.locations_config is not None:
+        with open(args.locations_config, mode="r", encoding="utf-8") as defs, open(
+            Path(args.release) / "local", mode="w", encoding="utf-8"
+        ) as local_activator, open(
+            Path(args.release) / "local.csh", mode="w", encoding="utf-8"
+        ) as local_csh_activator:
+            defs = yml.safe_load(defs)
+            local.write_local_activators(
+                data, defs, local_activator, local_csh_activator
+            )
 
     releasedoc = Path(args.release) / Path(args.release)
     with open(releasedoc, "w", encoding="utf-8") as filehandle:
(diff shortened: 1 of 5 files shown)


hnformentin:
Would it be possible to add tests for this new behaviour?

komodo/cli.py
```diff
@@ -88,13 +88,16 @@ def _main(args):
...
 
-    with open(args.locations_config, mode="r", encoding="utf-8") as defs, open(
-        Path(args.release) / "local", mode="w", encoding="utf-8"
-    ) as local_activator, open(
-        Path(args.release) / "local.csh", mode="w", encoding="utf-8"
-    ) as local_csh_activator:
-        defs = yml.safe_load(defs)
-        local.write_local_activators(data, defs, local_activator, local_csh_activator)
+    if args.locations_config is not None:
+        with open(args.locations_config, mode="r", encoding="utf-8") as defs, open(
```
berland, line 92: This  `with` construct is just so hard to read, when all it is doing is to read one file and write two others. But maybe a rewrite can be done in a different PR.

kwinkunks: I agree... TBH I didn't dare touch it and didn't want to dig into `komodo.local`. I made an issue for it.

komodo/cli.py
```diff
@@ -244,6 +241,12 @@ def cli_main():
     optional_args.add_argument("--postinst", "-P", type=str)
 
     optional_args.add_argument("--renamer", "-R", default="rename", type=str)
+    optional_args.add_argument(
+        "--locations-config",
+        type=str,
+        required=False,
+        help="Path to locations.yml, a map of location to a server.",
```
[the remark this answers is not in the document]
kwinkunks: Good one, done.

berland:
Should we have a integration test that certifies that no other part of the code assumes that there is a location config?

kwinkunks:
Good points about testing. Added a test to ensure that we can run `kmd` without the `--locations-config` argument and that it does not produce a `local` script.

berland (approved):
looking good!

kwinkunks:
:thinking: Hm, tests pass on local but fail on CI. Investigating...

kwinkunks:
:information_source: the problem was that `virtualenv` was not in the environment. Rather than trying to fix this, possibly getting into `bootstrap.sh` and other magic, I instead made my test more like the existing test in `test_cli.py`. This is okay, but has the disadvantage of having to add one more argument (`--extra-data-dirs`) to the "minimal" example.

PR #320, opened 2023-01-23
Add command-line features to symlink suggester

berland:
Making it possible to be more verbose and to do dry-runs.

Bonus: Pylinting, partial typing and py2 cleanup


komodo/symlink/suggester/cli.py
```diff
@@ -57,49 +61,64 @@ def _parse_args():
     parser.add_argument("--git-fork", help="git fork", default="equinor")
     parser.add_argument("--git-repo", help="git repo", default="komodo-releases")
     parser.add_argument("--git-ref", help="git ref", default="main")
+    parser.add_argument(
```
kwinkunks, line 64: Suggest adding `argparse.ArgumentDefaultsHelpFormatter` to this parser --- I added it to all of the other CLIs last week, but missed this one. (That other PR is not merged yet so I can also add it there, in which case I think I'll wait till this one is merged.)

komodo/symlink/suggester/cli.py
```diff
@@ -111,16 +130,28 @@ def suggest_symlink_configuration(args, repo):
...
 
-    pull = suggest_symlink_configuration(args, repo)
-    if pull is not None:
-        print(pull.html_url)
+    if args.verbose:
+        logger.setLevel(logging.INFO)
+
+    pull = suggest_symlink_configuration(args, repo, dry_run=args.dry_run)
+    if pull is None:
+        sys.exit(0)
```
hnformentin, line 152: What is this case for?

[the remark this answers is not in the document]
kwinkunks: FWIW I quite like `sys.exit()` when it's an expected, legitimate way to exit the program. In other words, it's not a "something went wrong" situation. (It is just another exception, after all.)

Admittedly it's sometimes hard to tell which is better, but to me this seems like a legit use case.

komodo/symlink/suggester/release.py
```diff
@@ -1,47 +1,47 @@
...
 
 
-class Release(object):
-    def __init__(self, release_id):
+class Release:
+    def __init__(self, release_id: str):
         self.release_id = release_id
 
-    def __repr__(self):
+    def __repr__(self) -> str:
```
kwinkunks, line 20: It's not always possible, but I think in general the `__repr__` should be something you could `eval`. So for example: `return f"Release({self.release_id})"`

The methods that want to use `repr(self)` etc seem like they could just use `len(self.release_id)`, which seems more explicit to me. Indeed, in `is_concrete()` I feel like a regex would be even more explicit.

IDK, not a big deal, just made me double-take.

komodo/symlink/suggester/cli.py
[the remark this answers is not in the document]
berland: I think it is ok to use `sys.exit() `in the main python file (`cli.py`) as this file is meant as a command line tool. In Python modules linked into this, there should be no `sys.exit()` (have not checked..)

komodo/symlink/suggester/cli.py
berland: I can do it here!

komodo/symlink/suggester/cli.py
kwinkunks: I guess `suggest_symlink_configuration()` returns `None` for a dry run (line 139), and if that comes back, we want to exit. Can't speak for Håvard but personally I think exiting from `main` is probably clearer than exiting from the other function.

berland: I have typed the code some more, and added a comment here!

berland: It is basically a legit condition where there is nothing to do.

komodo/symlink/suggester/release.py
berland: 🤔 Should the type of the returned object be a Callable?

berland: Maybe just leave it untyped

kwinkunks: Should be `str`

kwinkunks (approved):
This looks good to me.

PR #325, opened 2023-01-24
remove 'travis' from integration tests

### LINKED ISSUE #272: Remove the string travis from CI
The usage of "travis" in the CI should be updated to reflect that this is not any longer run on travis, but still relevant.

https://github.com/equinor/komodo/blob/1852315c1b80ff8b61996ad249c37a568adc4fc5/.github/workflows/test.yml#L50

kwinkunks:
Still checking that there aren't other tools/tests in the chain that used these scripts.

resolves #272


.github/workflows/test.yml
```diff
@@ -45,31 +45,31 @@ jobs:
...
             --prefix /tmp/pfx \
-            --release travis \
-            --locations-config $(realpath ci/travis/locations.yml) \
+            --release ci \
+            --locations-config $(realpath ci/locations.yml) \
 
     - name: Build libkmd.so for testing
       run: |
-        ci/travis/build_lib.sh /tmp/pfx/travis
+        ci/build_lib.sh /tmp/pfx/ci
```
berland, line 58: Use something under `${{runner.temp}}` instead of `/tmp` ?

.github/workflows/test.yml
```diff
@@ -45,31 +45,31 @@ jobs:
...
+            ci/repository.yml \
             --workspace /tmp/kmd-ws \
             --cache /tmp/kmd-cache \
             --prefix /tmp/pfx \
-            --release travis \
-            --locations-config $(realpath ci/travis/locations.yml) \
+            --release ci \
+            --locations-config $(realpath ci/locations.yml) \
 
     - name: Build libkmd.so for testing
```
berland, line 56: We might add "(a dummy C library)" to avoid people believing `libkmd` is something useful..

berland (approved):
🚀

PR #328, opened 2023-01-26
Deprecate pyver option in CLI

### LINKED ISSUE #322: Deprecate then remove pyver argument from kmd cli
This argument is not used in the code and should be deprecated (if anyone is passing it, removing it will break their code with `unrecognized arguments`).

kwinkunks:
Deprecated and testing but the warning is not enough to interrupt what the user sees when running the CLI. So I need to add something else.

Resolves #322

### DIFF
--- a/komodo/cli.py
+++ b/komodo/cli.py
@@ -1,7 +1,9 @@
 import argparse
 import os
 import sys
+import warnings
 from pathlib import Path
+from typing import List
 
 import jinja2
 import yaml as yml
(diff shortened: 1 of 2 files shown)


komodo/cli.py
```diff
@@ -340,13 +362,16 @@ def cli_main():
...
+    args = parser.parse_args(args)
 
-    if args.workspace and not Path(args.workspace).exists():
-        Path(args.workspace).mkdir()
+    if args.pyver is not None:
+        message = (
+            "\n⚠️  The --pyver option is deprecated and will be removed in a "
+            "future version of komodo. It is not used by komodo.\n"
+        )
+        print(message, file=sys.stderr)
```
berland, line 372: You could use a `FutureWarning` for this; https://docs.python.org/3/library/exceptions.html#FutureWarning

kwinkunks: Ok, that is much better :raised_hands: I assumed it would have the same problem as `DeprecationWarning` but no, it shows up just fine. I guess it must be [the default warning filters](https://docs.python.org/3/library/warnings.html#default-warning-filter). TIL.

hnformentin (approved):
LGTM and it seems you have addressed the comments of @berland :-)

PR #336, opened 2023-02-08
Let komodo-show-version print to stdout

berland:
Letting main() return a string will have it printed to stderr

### DIFF
--- a/komodo/show_version.py
+++ b/komodo/show_version.py
@@ -171,13 +171,14 @@ def parse_args(args: List[str]) -> argparse.Namespace:
     return parser.parse_args(args)
 
 
-def main() -> str:
+def main() -> int:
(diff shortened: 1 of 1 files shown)


kwinkunks (approved):
Nice one Håvard. I'm never sure how to emit the results from CLIs so will likely refer back to this some day. Wonder if you could explicitly print to `stdout` too? Anyway, good one.

berland:
> Nice one Håvard. I'm never sure how to emit the results from CLIs so will likely refer back to this some day. Wonder if you could explicitly print to `stdout` too? Anyway, good one.

You can always open the stdout descriptor as a file handle I suppose and write to it, but it would make it not so pretty. `print()`  always(?) default to stdout.

PR #338, opened 2023-02-09
Ignore only UnicodeDecodeError in shebang-fixing

### LINKED ISSUE #337: Handle exceptions in shebang_fix better
A log output from building komodo can be like this:

```
Fixup #! in pip-provided packages if bin exist
[/work/projects/komodo]> pip install treelib==1.6.1 --prefix builds/stable-0/root --no-index --no-deps --ignore-installed --cache-dir pip-cache --find-links pip-cache
b'Looking in links: pip-cache\nCollecting treelib==1.6.1\n  Using cached treelib-1.6.1-py3-none-any.whl\nInstalling collected packages: treelib\nSuccessfully installed treelib-1.6.1\n'
[/work/projects/komodo]> pip install pyscal==v0.10.2 --prefix builds/stable-0/root --no-index --no-deps --ignore-installed --cache-dir pip-cache --find-links pip-cache
b'Looking in links: pip-cache\nProcessing ./pip-cache/pyscal-0.10.2-py2.py3-none-any.whl\nInstalling collected packages: pyscal\nSuccessfully installed pyscal-0.10.2\n'
Exception in reading bin python: 'utf-8' codec can't decode byte 0xa8 in position 40: invalid start byte
Exception in reading bin python3: 'utf-8' codec can't decode byte 0xa8 in position 40: invalid start byte
Exception in reading bin python3.8: 'utf-8' codec can't decode byte 0xa8 in position 40: invalid start byte
```

The print at the first line here is badly placed, and also looks suspicious. We might skip that print altogether.

Then the exceptions should not be printed, as this is expected when we try to read a binary file as text.

berland:
Resolves #337

### DIFF
--- a/komodo/cli.py
+++ b/komodo/cli.py
@@ -141,7 +141,6 @@ def _main(args):
         # cases falling back to /tmp, which is undesired when building on nfs.
         os.environ["TMPDIR"] = args.tmp
 
-    print("Fixup #! in pip-provided packages if bin exist")
     release_path = Path(args.prefix) / Path(args.release)
     release_root = release_path / "root"
     for pkg, ver in args.pkgs.items():
--- a/komodo/shebang.py
+++ b/komodo/shebang.py
@@ -12,7 +12,7 @@ def fixup_python_shebangs(prefix, release):
     """Fix shebang to $PREFIX/bin/python.
 
     Some packages installed with pip do not respect target executable, that is,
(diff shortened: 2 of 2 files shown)


berland:
We should rather let it crash if an unexpected exception occurs, and then investigate and fix properly. That is why pylint doesn't like catching bare exceptions.

hnformentin (approved):

PR #339, opened 2023-02-09
Avoid certain combination of letters

berland:
This will save Norwegians many seconds a day not having to think twice on what this string really means.


kwinkunks (approved):
Seems reasonable. And now I'm wondering what kind of word runkmd looks like in Norwegian...

PR #340, opened 2023-02-10
Add ask for tags script

valentin-krasontovitsch:
The idea of this script is to automate part of the release management process - it allows checking for internal dependencies that have changes on their default branch since their latest release, and composes a message that can be used to poke maintainers and invite them to create a new tag.


oysteoh:
As discussed Valentin - this PR needs to wait until the release process and proposal of changing it has been approved in Scout

komodo/ask_for_tags.py
```diff
@@ -0,0 +1,218 @@
...
+from github.Repository import Repository
+
+not_in_komodo_releases_project_msg = (
+    "We expect this script to be run in the komodo-releases project"
+)
+
+
+def get_packages_that_use_main() -> List[str]:
+    bleeding_release_file = "releases/matrices/bleeding.yml"
+    duplicate_pkgs = ["everest-models"]  # this is included in the everest repo
```
kwinkunks, line 23: In principle, `komodo` is a general purpose FOSS tool. I wonder if it's possible to write the functions in this module in such a way that they do not assume it's only being run by Equinor? I will mark some other lines that can perhaps be refactored.

komodo/ask_for_tags.py
```diff
@@ -0,0 +1,218 @@
...
+    return main_pkgs_in_bleeding
+
+
+def get_repos(gh_client: Github, packages: List[str]) -> List[Repository]:
+    """tries to download repositories based on provided package names, using some
+    predefined fork names. throws if repo for a package could not be found / accessed
+    """
+
+    DEFAULT_FORK = "equinor"
+    ALTERNATE_FORK = "tno-everest"
```
kwinkunks, line 44: More Equinor-specific stuff.

komodo/ask_for_tags.py
```diff
@@ -0,0 +1,218 @@
...
+        release_commit = get_commit_of_latest_release(repo)
+        if last_commit != release_commit:
+            repos_with_changes.append(repo)
+    return repos_with_changes
+
+
+def get_scout_maintainers(gh_client: Github) -> Mapping[str, str]:
+    scout_repo = gh_client.get_repo("equinor/scout")
+    projects_file = scout_repo.get_contents("projects.md")
+    projects_file_contents = base64.b64decode(projects_file.content).decode()
```
kwinkunks, line 97: More Equinor specific stuff. Maybe we can generalize this, I'm not sure. We've been wondering if other parts of the code could benefit from a local config file -- possibly this is `locations.yaml` or maybe some new thing. This seems like the kind of thing that could go in there.

komodo/ask_for_tags.py
```diff
@@ -0,0 +1,218 @@
...
+    return "\n".join([intro_text, repo_with_maintainers_line, outro])
+
+
+def main():
+    github_token = os.getenv("GITHUB_TOKEN", None)
+    if github_token is None:
+        raise ValueError(
+            "Please provide a github access token through the env var `GITHUB_TOKEN`\n"
+            "Note that the token should be a personal access token "
+            "(not fine-grained),\n configured for SSO with equinor"
```
kwinkunks, line 198: More Equinor-specific stuff.

komodo/ask_for_tags.py
kwinkunks: (Just to add: another possibility is that this project is not meant for general consumption at all, which is totally fine -- but then we should either close it or make it clear that it's an Equinor tool.)

oysteoh:
> thanks for the comments @kwinkunks - i was thinking along those same lines, and of course you are right, this violates 12-factor in so many ways : )
> 
> i also now discovered a bug, and will try to address the points you raise while working on a fix!

I would also suggest to wait with the changes and pause work on this PR until we have a decision on how we want to proceed with the release process.

valentin-krasontovitsch:
i'm unassigning myself as I'm leaving scout - @eivindjahren, since we talked a lot about this, maybe you could try to find somebody to take this further (e.g. current up and coming release manager @yngve-sk), or close as not desired

jonathan-eq:
@eivindjahren Do you think we should close this as abandoned?

eivindjahren:
@jonathan-eq I think this is still up for discussion,

andreas-el:
Closed since this was implemented in Komodo-releases instead utilising this as the basis. Thanks @valentin-krasontovitsch ! ☀️

PR #344, opened 2023-02-20
Ignore invalid versions/specifiers in pypi checker

valentin-krasontovitsch:
As far as I understand, we pull all builds of a package from pypi, and then try to parse their version and python requirement. Some packages seem to have unparsable pre-release versions (absl-py: `1.3.0.dev-r634`), or invalid python specifiers (pyzmq: `!=3.0*`).

We ignore all types of such errors, presuming that they occur on builds that are irrelevant for us.

### DIFF
--- a/komodo/check_up_to_date_pypi.py
+++ b/komodo/check_up_to_date_pypi.py
@@ -6,7 +6,7 @@ import sys
 import requests
 import ruamel.yaml
 from packaging import version as get_version
-from packaging.specifiers import SpecifierSet
+from packaging.specifiers import InvalidSpecifier, SpecifierSet
 
 from komodo.package_version import LATEST_PACKAGE_ALIAS, strip_version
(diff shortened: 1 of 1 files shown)


DanSava:
> We ignore all types of such errors, presuming that they occur on builds that are irrelevant for us.

This sounds like a very strong assumption to make. Not saying it is wrong, just maybe double-check with somebody with a bit more experience with the komodo codebase.

valentin-krasontovitsch:
i have requested @berland and @oyvindeide - hoping that one of them will be able to say if this is OK : )

berland:
There is a linter in komodo that at least gives warnings for "dangerous" versions, I suppose these should be flagged like that as well.

berland:
Getting there, thanks for the context. I think this looks good!

berland (approved):

PR #346, opened 2023-02-22
Replace examples with doc-example

berland:
With this PR the example is actually buildable.


hnformentin:
Could you check the tests? They are failing in linting so far...

berland:
> Could you check the tests? They are failing in linting so far...

yup, they need an update..

berland:
> Could you check the tests? They are failing in linting so far...

Fixed!

hnformentin (approved):
LGTM!

PR #349, opened 2023-02-23
Recognize main

berland:


hnformentin (approved):
Good catch @berland 
Unfortunately there is no test for failure in linting, the nearest test I found was `test_lint` and it only tests a valid file. Do you think it would be worth to have an issue to add a test for these errors?

berland:
These are only warnings, not errors (even though they are seemingly called errors in the source), so I don't think I will add tests for them (now).

Issue #294, opened 2023-01-09
Rename the cache/dlprefix/outdir/output arguments


kwinkunks:
The same concept, a temporary directory for `pip` and other build tools to download into, has several names:

- `cache` in `cli.py` (and therefore in the `kmd` CLI)
- `dlprefix` in `build.py`
- `outdir` in komodo.fetch.fetch()
- `output` in the CLI implemented by komodo.fetch.py.

So there are at least 4 names for the same thing. This is confusing for the developer, but could lead to confusion for the user too, for example if they pass `cache` to the CLI, then it complains of a problem with `outdir` emitted by `fetch.py`.

It's going to need a `DeprecationWarning` because who knows who/what is using these various functions.

PR #351, opened 2023-02-23
Change command line option name from cache to work

berland:
The command line option cache hints to komodo supporting some kind of caching, which it does not. The directory named used as a cache is a working directory used while building, and cannot be prepopulated.

### DIFF
--- a/docs/basic-usage.md
+++ b/docs/basic-usage.md
@@ -52,7 +52,7 @@ A full software distribution can then be built and deployed to a specified
 path, e.g. `./builds/stable-0.0.1`, with the following command:
 
 ```bash
-kmd stable.yml repository.yml --prefix builds --release stable-0.0.1 --locations-config locations.yml --cache pip-cache
+kmd stable.yml repository.yml --prefix builds --release stable-0.0.1 --locations-config locations.yml
 ```
 
 To use this environment, type `source builds/stable-0.0.1/enable`.
--- a/komodo/cli.py
+++ b/komodo/cli.py
@@ -54,7 +54,7 @@ def _main(args):
     data = Data(extra_data_dirs=args.extra_data_dirs)
 
     if args.download or (not args.build and not args.install):
-        git_hashes = fetch(args.pkgs, args.repo, outdir=args.cache, pip=args.pip)
+        git_hashes = fetch(args.pkgs, args.repo, outdir=args.downloads, pip=args.pip)
 
     if args.download and not args.build:
         sys.exit(0)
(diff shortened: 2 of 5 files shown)


komodo/fetch.py
```diff
@@ -70,8 +70,13 @@ def fetch(pkgs, repo, outdir, pip="pip") -> dict:
...
         raise ValueError(
-            "The value of `outdir`, the cache location for pip and other "
-            "tools, cannot be None or the empty string."
+            "The value of `outdir`, the download destination location "
+            "cannot be None or the empty string."
+        )
+    if os.path.exists(outdir) and os.listdir(outdir):
+        print(
+            f"WARNING: Downloading to non-empty directory {outdir}, "
+            "this is not recommended."
```
kwinkunks, line 79: Actually, I think it's currently not _possible_, is it? I think the build will fail.

docs/basic-usage.md
```diff
@@ -52,7 +52,7 @@ A full software distribution can then be built and deployed to a specified
 path, e.g. `./builds/stable-0.0.1`, with the following command:
 
 ```bash
-kmd stable.yml repository.yml --prefix builds --release stable-0.0.1 --locations-config locations.yml --cache pip-cache
+kmd stable.yml repository.yml --prefix builds --release stable-0.0.1 --locations-config locations.yml
```
kwinkunks, line 55: Good catch!

komodo/cli.py
```diff
@@ -265,11 +268,14 @@ def parse_args(args: List[str]) -> argparse.Namespace:
...
-        "--cache",
-        "-c",
+        "--workspace_for_downloads",
+        "--cache",  # deprecated
+        "-c",  # deprecated
         type=str,
-        default="pip-cache",
-        help="The temporary directory used for downloads, e.g. by pip.",
+        default="_work",
+        help="A temporary destination directory used for downloads, "
```
kwinkunks, line 276: Is it temporary? I didn't catch where it is removed.

kwinkunks:
Good stuff -- this is much needed!

Wondering if this PR should just take on #294 ? I wasted a bunch of time figuring it out a few weeks ago (and failing!) and I'm guessing you (and Julius/Zohar I think) just burnt a bunch more time figuring it all out again. Maybe a sign that we can no longer afford this particular bit of debt.

I guess I'm especially wondering if it's a good idea to introduce more words (`work` and `workspace_for_downloads`) especially as we already have `workspace`, which is something else again? Maybe `download_directory` is clearer?

komodo/fetch.py
berland: At least it will fail if you use the protocol "nfs". Probably not for all other protocols, but behaviour is undefined. We should probably fail hard, but then we are at a breaking change.

berland:
Yes. I even discovered the "workspace" option after having written this. "workspace" could possibly be the only allowed argument, and we can require it to be empty.

kwinkunks (approved):
I like this, much cleaner.

kwinkunks:
Oh, I guess style is failing, but I'm sure you'll fix that :D

PR #352, opened 2023-02-23
Declare doc building dependency

berland:


hnformentin:
Is there a particular reason why @kwinkunks should be the reviewer?

kwinkunks (approved):
Good catch Håvard

berland:
> Is there a particular reason why @kwinkunks should be the reviewer?

No anyone can review, @kwinkunks just got an extra ping, having introduced the bug :)

PR #355, opened 2023-02-28
Make komodo-clean-repository able to reformat again

### LINKED ISSUE #354: Running komodo-clean-repository without check does not prettify yaml
In 82c43a83c39f2d728613d9bf8274b848a9d9dc4f the logic of the prettifyer was changed so that running without --check does nothing:
https://github.com/equinor/komodo/blob/d9b2344f24874b11b5e7cd75fb06dbfd3d1c50f3/komodo/release_cleanup.py#L183-L188

berland:
Resolves #354

### DIFF
--- a/komodo/release_cleanup.py
+++ b/komodo/release_cleanup.py
@@ -127,6 +127,7 @@ def add_prettier_parser(subparsers):
         nargs="+",
     )
     prettier_parser.add_argument(
+        "--check-only",
         "--check",
         action="store_true",
         help=(
@@ -180,9 +181,16 @@ def run_cleanup(args, parser):
 def run_prettier(args, _):
     release_files = [filename for sublist in args.files for filename in sublist]
 
-    if not args.check or all(
-        [prettified_yaml(filename, args.check) for filename in release_files]
+    if all(
+        [
+            prettified_yaml(filename, check_only=args.check_only)
+            for filename in release_files
+        ]
     ):
+        sys.exit(0)
+
+    if args.check_only is False:
+        # The prettifier has successfully reformatted all files
         sys.exit(0)
 
     sys.exit(1)
--- a/tests/test_release_cleanup.py
+++ b/tests/test_release_cleanup.py
@@ -1,4 +1,5 @@
 import os
+from pathlib import Path
 
 import pytest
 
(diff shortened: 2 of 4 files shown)


komodo/release_cleanup.py
```diff
@@ -180,11 +180,13 @@ def run_cleanup(args, parser):
...
 
-    if not args.check or all(
-        [prettified_yaml(filename, args.check) for filename in release_files]
+    if all(
+        [prettified_yaml(filename, check_only=args.check) for filename in release_files]
     ):
         sys.exit(0)
-
+    if args.check is False:
+        # The prettifier has successfully reformatted all files
```
hnformentin, line 188: Don't we need a step here to prettify?

tests/test_release_cleanup.py
```diff
@@ -108,6 +109,37 @@ def test_write_to_file(tmpdir):
...
+    with pytest.raises(SystemExit) as exit_info:
+        main(["prettier", "--files", repo_with_whitespace])
+        assert exit_info.value.code == 0
+
+    # IF checked again, the file should be ok, proving it was reformatted:
+    with pytest.raises(SystemExit) as exit_info:
+        main(["prettier", "--check", "--files", repo_with_whitespace])
+        assert exit_info.value.code == 0
+
+
```
hnformentin, line 142: Very good and clear test. Do we really need the capital letters in some words in comments?

komodo/release_cleanup.py
berland: Added a comment to make it a little bit clearer. I have thought about refactoring to increase readability but did not come up with a good (/easy) solution.

tests/test_release_cleanup.py
berland: It is an attempt at writing test comments according to what @eivindjahren has advocated in the Scout slack channel, on "behavioural driven development". The casing of initial words might not be important.

berland: Ah, that issue https://github.com/equinor/ert/pull/4429 was closed unmerged.

komodo/release_cleanup.py
berland: Change the name of the argument from `check` to `check-only`. I think that improves readability.

tests/test_release_cleanup.py
berland: Upper case removed.

berland:
Rebased and partially squashed. Last chance to review the git log :)

komodo/release_cleanup.py
hnformentin: Do you know if this imply any need of `deprecation warning`?

hnformentin:
> Rebased and partially squashed. Last chance to review the git log :)

LGTM also, please consider/answer my reply on `--check` before merging 🎉

komodo/release_cleanup.py
berland: I chose not to, since it would be an API break that might not be very important to do. I am not sure if we within argparse are able to deprecate option variants.

PR #356, opened 2023-03-01
Add time profiling

berland:


kwinkunks (commented):
Wondering a bit if there are any patterns for this kind of thing that would avoid scattering the timing code through the business logic? It's not something I've ever looked at before; I guess it's a kind of logging. I don't know what to suggest but, for example, could we make a decorator for the various functions `fetch`, `make`, `shell` etc instead?

berland:
Some hints might be taken from here: https://stackoverflow.com/questions/33987060/python-context-manager-that-measures-time

xjules (approved):
Nice! :rocket:

berland:
Decided to merge it even if it is at a primitive stage for the sake of data gathering.

PR #363, opened 2023-04-26
Make suggester branches more unique

xjules:
This hopefully makes the creation of (azure) branch more stable.

### DIFF
--- a/komodo/symlink/suggester/cli.py
+++ b/komodo/symlink/suggester/cli.py
@@ -92,9 +92,10 @@ def suggest_symlink_configuration(
         logger.info("Nothing to update")
         return None
 
-    target_branch = f"{args.release}/{args.mode}"
     if "azure" in args.symlink_conf_path:
-        target_branch += "/azure"
+        target_branch = f"{args.release}/azure-{args.mode}"
+    else:
+        target_branch = f"{args.release}/onprem-{args.mode}"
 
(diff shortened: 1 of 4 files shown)


komodo/symlink/suggester/cli.py
```diff
@@ -94,7 +94,7 @@ def suggest_symlink_configuration(
 
     target_branch = f"{args.release}/{args.mode}"
     if "azure" in args.symlink_conf_path:
-        target_branch += "/azure"
+        target_branch = f"{args.release}/azure-{args.mode}"
```
berland, line 97: What about treating onprem and azure equivalently, so that `target_branch` contains either `onprem-stable` or `azure-stable`?

xjules:
I need to fix the `bootstrap.sh` also - I guess I will do it in this PR.

berland:
I think I would like the bootstrap things in a separate commit

berland:
Do we know why the test workflow started to fail, spurring the fixed in bootstrap?

xjules:
> Do we know why the test workflow started to fail, spurring the fixed in bootstrap?

My best guess is that is had been working just by pure luck as the pip used in pytests (in  test_cli_main) was never specified and it was using the default system pip.

berland (approved):
👍🏻

PR #367, opened 2023-05-09
Remove unused jenkins/link.sh

berland:
No usage found. 

File has not been updated since 2019.


kwinkunks (approved):
Looks like some ancient symbolic link creator? Good cleanup.

PR #370, opened 2023-05-09
Merge build_lib.sh directly into workflow yml file

berland:


kwinkunks (approved):
I guess I can't quite visualize how this test could fail, but at least it does indeed seem to do what it did before. Definitely great to remove this extra script.

PR #376, opened 2023-07-10
Make creating symlinks verbose

berland:
This allows for forwarding of what the symlink creator actually did to status dashboards.

### DIFF
--- a/komodo/symlink/create_links.py
+++ b/komodo/symlink/create_links.py
@@ -25,9 +25,13 @@ def _create_link(src, dst, link_dict):
         raise ValueError(f"{src} does not exist")
 
     if os.path.exists(dst) and os.path.islink(dst):
-        if os.readlink(dst) == src:
+        existing_link = os.readlink(dst)
+        if existing_link == src:
             return
         os.remove(dst)
+        print(f"Existing symlink {dst} moved from {existing_link} to {src}")
+    else:
+        print(f"Created new symlink {dst} pointing to {src}")
 
     os.symlink(src, dst)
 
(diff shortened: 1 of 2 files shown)


berland:
This should have a follow-up PR in komodo-releases to update the slack notification to capture this output and only send a slack message if this output is nonzero (and then use this output as the message content)

hnformentin (commented):
LGTM...but what do you think about creating an issue for the coming up PR?

xjules:
Looks very good, although it is quite easy to get the config file that was changed from the push action (in `update_symlink.yml`):
```yml
- id: extract_file_changed
          run: |
            for config_file in $(git diff --name-only ${{ github.event.before }} ${{ github.event.after }}); do
              echo "The following file was modified $config_file"
              echo "config_file=$config_file" >> $GITHUB_OUTPUT
            done
```
which I've already started writing, but this looks to be more solid approach.

hnformentin:
Seeing the two suggestions in perspective, I tend to think that this one is cleaner than the one proposed in https://github.com/equinor/komodo/pull/376.
Can we wait to @berland to be back before merging?

valentin-krasontovitsch (approved):
looks good!

It seems to me that we need both this PR and [komodo-release/4112](https://github.com/equinor/komodo-releases/pull/4112/files) [INTERNAL LINK] - this PR provides the output that can be used e.g. in a slack message, but the latter PR corrects on which infrastructure the update symlink job is run.

do you concur, @hnformentin @xjules ? if you agree, i think we can merge both

PR #381, opened 2023-08-14
Add input validation to transpiler and combine

### LINKED ISSUE #362: Add validation to inputs of transpiler and combine
A couple of errors/failures could be avoided if we had validation of the inputs for those functions, for example:
- [having a new line as release name](https://github.com/equinor/komodo-releases/actions/runs/4724626459/jobs/8382076293)
- [this PR would also benefit from that](https://github.com/equinor/komodo-releases/pull/3507)

A suggestion is to validate the inputs in the argparse schema with validation via `type` property.
For example, from ert:

>     ensemble_experiment_parser.add_argument(
>         "--current-case",
>         type=valid_case,
>         default="default",
>         help="Name of the case where the results for the simulation "
>         "using the prior parameters will be stored.",
>     ) 

where:

> def valid_case(user_input: str) -> Union[str, UUID]:
>     if user_input.startswith("UUID="):
>         return UUID(user_input[5:])
>     return valid_name(user_input)

In this case, `valid_case` is a function that validates the input and returns the correct data type.

jonathan-eq:
Resolves #362

### DIFF
--- a/komodo/cli.py
+++ b/komodo/cli.py
@@ -103,10 +103,10 @@ def _main(args):
         _print_timings(timings[-1])
 
         shell(f"mv {args.release + str(tmp_prefix)} {args.release}")
-        shell(
-            "rmdir -p --ignore-fail-on-non-empty "
-            f"{args.release + str(tmp_prefix.parent)}"
-        )
+        try:
+            os.removedirs(f"{args.release + str(tmp_prefix.parent)}")
+        except OSError:
+            pass
 
     if args.build and not args.install:
         sys.exit(0)
--- a/komodo/matrix.py
+++ b/komodo/matrix.py
@@ -14,7 +14,7 @@ def get_matrix(
     current release matrix."""
     for product in itertools.product(rhel_versions, py_versions):
         rh_ver, py_ver = product
-        yield (f"rhel{rh_ver}", f"py{py_ver.replace('.', '')}")
+        yield (f"rhel{rh_ver}", f"py{str(py_ver).replace('.', '')}")
 
 
 def format_release(base: str, rhel_ver: str, py_ver: str) -> str:
--- a/komodo/release_transpiler.py
+++ b/komodo/release_transpiler.py
@@ -3,10 +3,10 @@
 import argparse
 import itertools
 import os
-import re
-from typing import Optional, Sequence
+from typing import List, Optional, Sequence
 
 import yaml
+from packaging import version as version_parser
 
 from komodo.matrix import format_release, get_matrix
 from komodo.prettier import load_yaml, write_to_file
@@ -204,20 +204,32 @@ Combine release files into a matrix file. Output format:
     matrix_parser.add_argument(
         "--release-folder",
         required=True,
+        type=dir_path,
         help="Folder with existing release file (default: None)",
     )
     matrix_parser.add_argument(
         "--override-mapping",
         required=True,
+        type=valid_file,
         help="File containing explicit matrix packages (default: None)",
     )
+
+    def comma_delimited_python_versions(python_versions: str) -> List[str]:
+        output_list: List[str] = []
+        for python_version in python_versions.split(","):
+            parsed_python_version = version_parser.parse(python_version).base_version
+            if parsed_python_version >= "4.0.0":
+                raise version_parser.InvalidVersion(python_version)
+            output_list.append(parsed_python_version)
+        return output_list
+
     matrix_parser.add_argument(
         "--py_coords",
-        help="""Comma delimitated list of python versions to be combined,
+        help="""Comma delimited list of python versions to be combined,
         for example, "3.6,3.8" (without spaces).
         If None, the release files in release-folder will be used to imply
         the versions to combine.""",
-        type=lambda s: re.split(",", s),
+        type=comma_delimited_python_versions,
         required=False,
         default=None,
     )
@@ -231,11 +243,13 @@ Combine release files into a matrix file. Output format:
     transpile_parser.add_argument(
         "--matrix-file",
         required=True,
+        type=valid_file,
         help="Yaml file describing the release matrix",
     )
     transpile_parser.add_argument(
         "--output-folder",
         required=True,
+        type=dir_path,
         help="Folder to output new release files",
     )
     transpile_parser.add_argument(
@@ -247,6 +261,18 @@ Combine release files into a matrix file. Output format:
     )
     args = parser.parse_args()
     args.func(args)
+
+
+def valid_file(path: str) -> str:
+    if os.path.isfile(path):
+        return path
+    raise FileNotFoundError(path)
+
+
+def dir_path(should_be_valid_path: str) -> str:
+    if os.path.isdir(should_be_valid_path):
+        return should_be_valid_path
+    raise NotADirectoryError(should_be_valid_path)
 
 
 if __name__ == "__main__":
--- a/tests/test_release_transpiler_argument_types.py
+++ b/tests/test_release_transpiler_argument_types.py
@@ -1,0 +1,557 @@
+import argparse
+import sys
+from contextlib import contextmanager
+from os.path import abspath, dirname
+from typing import List
+
+import pytest
+
+from komodo.release_transpiler import main as release_transpiler_main
+
+VALID_RELEASE_FOLDER = abspath(dirname(__file__) + "/data/test_releases")
+VALID_RELEASE_BASE = "2020.01.a1"
+VALID_OVERRIDE_MAPPING_FILE = abspath(
+    dirname(dirname(__file__)) + "/examples/stable.yml"
+)
+VALID_PYTHON_COORDS = "3.6,3.8"
+VALID_MATRIX_FILE = abspath(dirname(__file__) + "/data/test_release_matrix.yml")
+VALID_OUTPUT_FOLDER = abspath(dirname(__file__))
+VALID_MATRIX_COORDINATES = "{rhel: ['7'], py: ['3.8']}"
+
+
+@contextmanager
+def does_not_raise():
+    yield
+
+
+@pytest.mark.parametrize(
+    "args, expectation",
+    [
+        (
+            [
+                "combine",
+                "--release-folder",
+                VALID_RELEASE_FOLDER,
+                "--release-base",
+                VALID_RELEASE_BASE,
+                "--override-mapping",
+                VALID_OVERRIDE_MAPPING_FILE,
+                "--py_coords",
+                "3.8,3.6",
+            ],
+            does_not_raise(),
+        ),
+        (
+            [
+                "combine",
+                "--release-folder",
+                VALID_RELEASE_FOLDER,
+                "--release-base",
+                VALID_RELEASE_BASE,
+                "--override-mapping",
+                VALID_OVERRIDE_MAPPING_FILE,
+            ],
+            does_not_raise(),
+        ),
+        (
+            [
+                "combine",
+                "--release-folder",
+                VALID_RELEASE_FOLDER,
+                "--release-base",
+                VALID_RELEASE_BASE,
+                "--override-mapping",
+                VALID_OVERRIDE_MAPPING_FILE,
+                "--py_coords",
+                "py3.8",
+            ],
+            pytest.raises(SystemExit),
+        ),
+        (
+            [
+                "combine",
+                "--release-folder",
+                VALID_RELEASE_FOLDER,
+                "--release-base",
+                VALID_RELEASE_BASE,
+                "--override-mapping",
+                VALID_OVERRIDE_MAPPING_FILE,
+                "--py_coords",
+                "test",
+            ],
+            pytest.raises(SystemExit),
+        ),
+        (
+            [
+                "combine",
+                "--release-folder",
+                VALID_RELEASE_FOLDER,
+                "--release-base",
+                VALID_RELEASE_BASE,
+                "--override-mapping",
+                VALID_OVERRIDE_MAPPING_FILE,
+                "--py_coords",
+                "'3.8, 3.6'",
+            ],
+            pytest.raises(SystemExit),
+        ),
+        (
+            [
+                "combine",
+                "--release-folder",
+                VALID_RELEASE_FOLDER,
+                "--release-base",
+                VALID_RELEASE_BASE,
+                "--override-mapping",
+                VALID_OVERRIDE_MAPPING_FILE,
+                "--py_coords",
+                "4.2",
+            ],
+            pytest.raises(SystemExit),
+        ),
+        (
+            [
+                "combine",
+                "--release-folder",
+                VALID_RELEASE_FOLDER,
+                "--release-base",
+                VALID_RELEASE_BASE,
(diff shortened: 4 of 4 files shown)


komodo/cli.py
```diff
@@ -103,10 +103,14 @@ def _main(args):
         _print_timings(timings[-1])
 
         shell(f"mv {args.release + str(tmp_prefix)} {args.release}")
-        shell(
-            "rmdir -p --ignore-fail-on-non-empty "
-            f"{args.release + str(tmp_prefix.parent)}"
-        )
+        if sys.platform == "darwin":
+            # OS X
+            shell("rm -rf " f"{args.release + str(tmp_prefix.parent)}")
```
berland, line 108: This would introduce different behaviour on OS X, as this will delete it always, while the rmdir version will let the directory stay there if it is nonempty.

berland: (or am I missing something?)

berland: Maybe `os.removedirs()` can be used instead for portability.

jonathan-eq: You are correct, --ignore-fail-on-non-empty is not the same as -f. I will look for another work around. Thank you 👍

komodo/matrix.py
```diff
@@ -14,7 +14,7 @@ def get_matrix(
     current release matrix."""
     for product in itertools.product(rhel_versions, py_versions):
         rh_ver, py_ver = product
-        yield (f"rhel{rh_ver}", f"py{py_ver.replace('.', '')}")
+        yield (f"rhel{rh_ver}", f"py{str(py_ver).replace('.', '')}")
```
berland, line 17: is `str()` necessary here? `py_versions` is typed as `Sequence[str]` and `mypy` should locate any abuses of that (not having checked to which degree we enforce mypy here)

komodo/release_transpiler.py
```diff
@@ -204,20 +204,31 @@ def main():
     matrix_parser.add_argument(
         "--release-folder",
         required=True,
+        type=dir_path,
         help="Folder with existing release file (default: None)",
     )
     matrix_parser.add_argument(
         "--override-mapping",
         required=True,
+        type=check_if_valid_file,
```
berland, line 213: since this is a "type", should we rename from `check_if_valid_file` to simply `valid_file`?

komodo/release_transpiler.py
```diff
@@ -204,20 +204,31 @@ def main():
...
+        output_list = []
+        for item in python_versions.split(","):
+            if re.match(r"^[2,3](\.\d+)?$", item) is None:
+                raise TypeError(item)
+            output_list.append(item)
+        return output_list
+
     matrix_parser.add_argument(
         "--py_coords",
         help="""Comma delimitated list of python versions to be combined,
```
berland, line 227: A typo that might be fixed in this PR: delimitated -> delimited

komodo/release_transpiler.py
```diff
@@ -204,20 +204,31 @@ def main():
...
         for example, "3.6,3.8" (without spaces).
         If None, the release files in release-folder will be used to imply
         the versions to combine.""",
-        type=lambda s: re.split(",", s),
+        type=checkValidPythonVersion,
```
berland, line 231: maybe rename to `valid_python_version_list`, avoiding CamelCase and in line with the suggestion above. 

It is also an alternative to go fully explicit: `comma_delimited_python_versions`

komodo/release_transpiler.py
```diff
@@ -204,20 +204,31 @@ def main():
...
         "--override-mapping",
         required=True,
+        type=check_if_valid_file,
         help="File containing explicit matrix packages (default: None)",
     )
+
+    def checkValidPythonVersion(python_versions: str) -> list:
+        output_list = []
+        for item in python_versions.split(","):
+            if re.match(r"^[2,3](\.\d+)?$", item) is None:
```
berland, line 220: If we just want to require a parseable version string, we might use `packaging.version.parse()` instead of maintaining our own regexp (doomed to fail for Python 4 :) )

komodo/release_transpiler.py
```diff
@@ -204,20 +204,31 @@ def main():
...
         help="Folder with existing release file (default: None)",
     )
     matrix_parser.add_argument(
         "--override-mapping",
         required=True,
+        type=check_if_valid_file,
         help="File containing explicit matrix packages (default: None)",
     )
+
+    def checkValidPythonVersion(python_versions: str) -> list:
```
berland, line 217: We can be more explicit in the return type: `list` can be changed to `List[str]`

tests/test_release_transpiler_argument_types.py
```diff
@@ -0,0 +1,330 @@
...
+VALID_RELEASE_BASE = "2020.01.a1"
+VALID_OVERRIDE_MAPPING_FILE = abspath(
+    dirname(dirname(__file__)) + "/examples/stable.yml"
+)  # "/Users/JONAK/Documents/FMU/SCOUT/komodo/examples/stable.yml"
+VALID_PYTHON_COORDS = "3.6,3.8"
+
+
+# TESTING COMBINER
+@pytest.mark.parametrize(
+    "args, expectedExitCode",
```
berland, line 19: we prefer to avoid camelCase :)

berland (changes requested):

tests/test_release_transpiler_argument_types.py
```diff
@@ -0,0 +1,330 @@
...
+                f"--release-folder {VALID_RELEASE_FOLDER} --release-base"
+                f" {VALID_RELEASE_BASE} --override-mapping"
+                f" {VALID_OVERRIDE_MAPPING_FILE} --py_coords true"
+            ),
+            512,
+        ),
+    ],
+)
+def test_combine_py_coords_type(args, expectedExitCode):
+    exit_code = os.system(f"{RELEASE_TRANSPILER_FILE_PATH} combine {args}")
```
berland, line 72: When running the test code, you can assume that komodo has been pip-installed. That means that the transpiler has already been installed in PATH (as defined in pyproject.toml, with the name `komodo-transpiler`) (then you don't need to fiddle around with the file path to the python file).

Another way this is often accomplished is as in `test_cli.py` where `sys.argv` is modified, and then the `main()` function in the transpiler file is called as a Python function. Then we don't' have the overhead of running the the transpiler in a separate process (and the code in `test_cli.py` would be even prettier if it used a mock for sys.argv, and not modifying it directly).

tests/test_release_transpiler_argument_types.py
```diff
@@ -0,0 +1,330 @@
...
+            ),
+            0,
+        ),
+        (
+            (
+                f"--release-folder {VALID_RELEASE_FOLDER} --release-base"
+                f" {VALID_RELEASE_BASE} --override-mapping"
+                f" {VALID_OVERRIDE_MAPPING_FILE} --py_coords py3.8"
+            ),
+            512,
```
berland, line 43: This magic number should probably be defined in a constant with a descriptive name.

tests/test_release_transpiler_argument_types.py
```diff
@@ -0,0 +1,330 @@
...
+import pytest
+
+RELEASE_TRANSPILER_FILE_PATH = abspath(
+    dirname(dirname(abspath(__file__))) + "/komodo/release_transpiler.py"
+)
+VALID_RELEASE_FOLDER = abspath(dirname(__file__) + "/data/test_releases")
+VALID_RELEASE_BASE = "2020.01.a1"
+VALID_OVERRIDE_MAPPING_FILE = abspath(
+    dirname(dirname(__file__)) + "/examples/stable.yml"
+)  # "/Users/JONAK/Documents/FMU/SCOUT/komodo/examples/stable.yml"
```
berland, line 13: rm comment

komodo/matrix.py
jonathan-eq: It threw an exception that the replace method did not exist on type float, so I casted it to make sure it was a string.

komodo/release_transpiler.py
jonathan-eq: Good idea!

komodo/release_transpiler.py
jonathan-eq: Good catch!

komodo/release_transpiler.py
jonathan-eq: I agree, the function name would be way more descriptive as comma_delimited_python_versions.

komodo/cli.py
```diff
@@ -103,10 +103,10 @@ def _main(args):
         _print_timings(timings[-1])
 
         shell(f"mv {args.release + str(tmp_prefix)} {args.release}")
-        shell(
-            "rmdir -p --ignore-fail-on-non-empty "
-            f"{args.release + str(tmp_prefix.parent)}"
-        )
+        try:
```
berland, line 106: This change should be in a separate commit, as it is motivated by mac os compatibility

(when we later look at the blame for this, we will not understand why we claimed that this was for input validation)

komodo/release_transpiler.py
```diff
@@ -249,5 +264,17 @@ def main():
     args.func(args)
 
 
+def valid_file(path: str) -> str:
+    if os.path.isfile(path):
+        return path
+    raise TypeError(path)
```
berland, line 270: This is not necessarily a TypeError, it is more likely it is a missing file. There is probably a FileNotFoundErrorException that can be used.

komodo/release_transpiler.py
```diff
@@ -249,5 +264,17 @@ def main():
...
+
+
+def dir_path(should_be_valid_path: str) -> str:
+    if os.path.isdir(should_be_valid_path):
+        return should_be_valid_path
+    raise TypeError(should_be_valid_path)
```
berland, line 276: ditto

tests/test_release_transpiler_argument_types.py
```diff
@@ -0,0 +1,564 @@
+import argparse
+import sys
+from contextlib import contextmanager
+from os.path import abspath, dirname
+
+import pytest
+
+from komodo.release_transpiler import main as release_transpiler_main
+
+RELEASE_TRANSPILER_FILE_PATH = abspath(
```
berland, line 10: this constant is probably unused?

tests/test_release_transpiler_argument_types.py
```diff
@@ -0,0 +1,564 @@
...
+                VALID_OVERRIDE_MAPPING_FILE,
+                "--py_coords",
+                "true",
+            ],
+            pytest.raises(ARGUMENT_PARSING_ERROR),
+        ),
+    ],
+)
+def test_combine_py_coords_type(args: list, expectation, monkeypatch):
+    print("args: " + " ".join(args))
```
berland, line 131: remove the print

tests/test_release_transpiler_argument_types.py
```diff
@@ -0,0 +1,564 @@
...
+                "--override-mapping",
+                VALID_OVERRIDE_MAPPING_FILE,
+                "--py_coords",
+                "true",
+            ],
+            pytest.raises(ARGUMENT_PARSING_ERROR),
+        ),
+    ],
+)
+def test_combine_py_coords_type(args: list, expectation, monkeypatch):
```
berland, line 130: We can use `List[str]` instead of only  `list` to be even more precise.

tests/test_release_transpiler_argument_types.py
```diff
@@ -0,0 +1,564 @@
...
+    ],
+)
+def test_combine_py_release_base_type(args: list, expectation, monkeypatch):
+    monkeypatch.setattr(sys, "argv", [""] + args)
+    with expectation:
+        release_transpiler_main()
+
+
+# TESTING TRANSPILER
+VALID_MATRIX_FILE = abspath(dirname(__file__) + "/data/test_release_matrix.yml")
```
berland, line 347: Can you run pylint on this file? I think it will complain that not all constants are defined at the top.

tests/test_release_transpiler_argument_types.py
```diff
@@ -0,0 +1,564 @@
...
+        ),
+    ],
+)
+def test_combine_py_release_base_type(args: list, expectation, monkeypatch):
+    monkeypatch.setattr(sys, "argv", [""] + args)
+    with expectation:
+        release_transpiler_main()
+
+
+# TESTING TRANSPILER
```
berland, line 346: I think this comment should go away, what we are really doing should be described by the function name (although sometimes we need to scroll down a bit to read that)

tests/test_release_transpiler_argument_types.py
jonathan-eq: Pylint did not mention it, but I will move it up regardless

tests/test_release_transpiler_argument_types.py
jonathan-eq: I got "sveiseblind" by looking at the same code for so long 👨‍🏭

tests/test_release_transpiler_argument_types.py
jonathan-eq: I was unsure about this one, I feel like main could be misunderstood and someone might assume main is the main function of the test file; not the release_transpiler_main.
Should I remove the constant and just use main instead?

komodo/cli.py
jonathan-eq: This was in the "Fix cli unit test not running on mac"-commit. Should I put it in another PR?
The commit name might be misleading, what would be a more suitable title?

komodo/release_transpiler.py
jonathan-eq: I will change this one to NotADirectoryError

tests/test_release_transpiler_argument_types.py
berland: It is the constant `RELEASE_TRANSPILER_FILE_PATH` that I guess is unused.

tests/test_release_transpiler_argument_types.py
```diff
@@ -0,0 +1,565 @@
...
+    dirname(dirname(abspath(__file__))) + "/komodo/release_transpiler.py"
+)
+VALID_RELEASE_FOLDER = abspath(dirname(__file__) + "/data/test_releases")
+VALID_RELEASE_BASE = "2020.01.a1"
+VALID_OVERRIDE_MAPPING_FILE = abspath(
+    dirname(dirname(__file__)) + "/examples/stable.yml"
+)
+VALID_PYTHON_COORDS = "3.6,3.8"
+
+ARGUMENT_PARSING_ERROR = SystemExit
```
berland, line 21: I think you can skip having these as constants, it should be just as readable in the test code to just use the exceptions as-is.

tests/test_release_transpiler_argument_types.py
jonathan-eq: Okay, I misunderstood. Is this something pylint should have caught?

berland (approved):
Looking very good!

berland:
I suggest rebasing manually into two commits, one with the mac-fix, and one commit with the combine/transpiler fix (the mac fix could have been its own PR since it does not solve issue 362, but that is not important as it will not matter for the git commit log)

PR #384, opened 2023-09-01
Improve yaml config file validation and linting

### LINKED ISSUE #377: Validate YAML config files as part of linting
Currently if I specify the version as (say) `3.1`, then attempting to build comes back with `missing version for networkx 3.1`, not something like `malformed version` or `you forgot to put quotes around it` -- or maybe even just deal with it because it should not matter. The problem is that the YAML parser mistakes `3.1` for a float, when we need it to be interpreted as a string. Two possible solutions I can think of:

- Use a YAML schema and validate the YAML 'properly'. Maybe this approach can also just coerce the version into a string.
- Get the linting to check for this mistake and report it more helpfully.

### LINKED ISSUE #382: Package case style causes problems
example: pyyaml vs PyYAML, now these are mismatches. If possible, we should make all the packages lower case, or at least make the lookup case insensitive, so that if:

Ex1: An upgrade proposal is made with pyyaml, and it is specified as PyYAML in repository.yaml, it should be detected accordingly.
Ex2: A release file has the package written as PyYAML, and an upgrade proposal has pyyaml, this introduces a "mismatch"

### LINKED ISSUE #383: `make` entry of `repository.yaml` should be validated earlier
Ref: https://github.com/equinor/komodo-releases/blob/fbfc58e28589628974a30b3d8feae19b637c2739/repository.yml#L17472
It should be checked that the `make` entry is valid, this currently shows up as an error a bit harder to decipher, ref: https://github.com/equinor/komodo-releases/actions/runs/6010649782/job/16303506165#step:8:1799

jonathan-eq:
Resolves #383 
Resolves #382
Resolves #377 
Resolves https://github.com/equinor/komodo-releases/issues/3362


komodo/yaml_file_types.py
```diff
@@ -0,0 +1,642 @@
...
+        super().__init__("r", *args, **kwargs)
+
+    def __call__(self, value):
+        file_handle = super().__call__(value)
+        yaml = YAML()
+        yml = yaml.load(file_handle)
+        file_handle.close()
+        assert isinstance(
+            yml, dict
+        ), "File could not be parsed as yaml. Make sure the formatting is correct"
```
berland, line 60: For nitpicking, since the yaml parser has not thrown an exception at this point, I guess it must be valid yaml. It has to be the wrong syntax with respect to komodo files though.

komodo/yaml_file_types.py
```diff
@@ -0,0 +1,642 @@
...
+MISSING_DEPENDENCY = "missing dependency"
+MISSING_MAINTAINER = "missing maintainer"
+MISSING_MAKE = "missing make information"
+MALFORMED_VERSION = "malformed version"
+MAIN_VERSION = "dangerous version (main branch)"
+MASTER_VERSION = "dangerous version (master branch)"
+FLOAT_VERSION = "dangerous version (float interpretable)"
+
+
+def _komodo_error(package=None, version=None, maintainer=None, depends=None, err=None):
```
berland, line 33: type the args?

komodo/yaml_file_types.py
```diff
@@ -0,0 +1,642 @@
...
+    Return the data from 'release' YAML file, but validate it first.
+    """
+
+    def __call__(self, value: str):
+        yml: dict = super().__call__(value)
+        self.validate_release_file(yml)
+        self.yml = yml
+        return self
+
+    def from_yaml_string(self, value: bytes):
```
berland, line 75: is  bytes the correct type here? I would guess it should be something similar to a filehandle. I have not checked whether mypy is enforced in the CI here.

berland:
Can you update the PR with information on which issue is being solved?

komodo/yaml_file_types.py
```diff
@@ -0,0 +1,642 @@
+import argparse
+import os
+from collections import namedtuple
+from pathlib import Path
+from typing import Dict, List
+
+from ruamel.yaml import YAML
+
+komodo_error = namedtuple(
```
berland, line 9: Is pylint confortable with these constants not being upper-case?

komodo/yaml_file_types.py
jonathan-eq: Yes, that makes sense! If it is not a yaml file, the super will already have raised an exception

berland:
The magnitude (lines of code and number of changed files) of this PR, and the big (more than one) number of issues it solves, warrants a split into multiple PR's. This will make the review process easier.

berland:
Is this PR still valid or should it be closed?

PR #386, opened 2023-09-05
Group upgrades komodo check pypi

jonathan-eq:
Resolves #365 
Should not depend on but,
Depends on #388

PR #387, opened 2023-09-05
Add flag for ignoring entries check pypi script

### LINKED ISSUE #342: add flag for ignoring entries with main in check pypi versions script

### LINKED ISSUE #414: Check pypi script not working with packages with only prerelease versions
Certain packages, for example "opentelemetry-semantic-conventions", only have prerelease versions. In this case, the script will try to run max() over an empty list and fail.

jonathan-eq:
Resolves #342
Resolves #414 

Depends on #386

### DIFF
--- a/komodo/check_up_to_date_pypi.py
+++ b/komodo/check_up_to_date_pypi.py
@@ -1,6 +1,7 @@
 import argparse
 import copy
 import pathlib
+import re
 import sys
 
 import requests
@@ -147,9 +148,17 @@ def run_check_up_to_date(
         f"{sys.version_info.micro}"
     ),
     propose_upgrade=False,
+    ignore=None,
 ):
     yaml = yaml_parser()
-    releases = load_from_file(yaml, release_file)
+    releases: dict = load_from_file(yaml, release_file)
+
+    if ignore:
+        releases = {
+            package_name: package_version
+            for package_name, package_version in releases.items()
+            if re.search(ignore, f"{package_name} {package_version}") is None
+        }
     repository = load_from_file(yaml, repository_file)
     upgrade_proposals_from_pypi = get_upgrade_proposals_from_pypi(
         releases,
@@ -161,7 +170,7 @@ def run_check_up_to_date(
             insert_upgrade_proposals(upgrade_proposals_from_pypi, repository, releases)
             print(
                 "Writing upgrade proposals from pypi, "
-                "assuming nothing has changed with dependencies...",
+                "assuming nothing has changed with dependencies..."
             )
             with open(propose_upgrade, mode="w", encoding="utf-8") as fout:
                 yaml.dump(releases, fout)
@@ -170,7 +179,11 @@ def run_check_up_to_date(
         major_upgrades, minor_upgrades, patch_upgrades, other_upgrades = [], [], [], []
         for name, versions in upgrade_proposals_from_pypi.items():
             pypi_latest = get_version.Version(versions["suggested"])
-            current_version = get_version.Version(versions["previous"])
+            try:
+                current_version = get_version.Version(versions["previous"])
+            except get_version.InvalidVersion:
+                print(f"Could not parse version {versions['previous']}")
+                continue
             if pypi_latest.major > current_version.major:
                 major_upgrades.append(
                     f"{name} not at latest pypi version: {pypi_latest}, "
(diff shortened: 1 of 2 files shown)


komodo/check_up_to_date_pypi.py
```diff
@@ -147,9 +148,14 @@ def run_check_up_to_date(
...
     propose_upgrade=False,
+    ignore=None,
 ):
     yaml = yaml_parser()
-    releases = load_from_file(yaml, release_file)
+    releases: dict = load_from_file(yaml, release_file)
+    if ignore:
+        for package_name, package_version in list(releases.items()):
+            if re.search(ignore, f"{package_name} {package_version}"):
+                del releases[package_name]
```
berland, line 158: An alternative way could be to do a dict comprehension, then you could avoid the [list(releases.items())](url) (it is usually an antipattern and some linters could erroneously suggest to remove the `list` function here)

komodo/check_up_to_date_pypi.py
```diff
@@ -161,16 +167,20 @@ def run_check_up_to_date(
             insert_upgrade_proposals(upgrade_proposals_from_pypi, repository, releases)
             print(
                 "Writing upgrade proposals from pypi, "
-                "assuming nothing has changed with dependencies...",
+                "assuming nothing has changed with dependencies..."
             )
-            with open(propose_upgrade, mode="w", encoding="utf-8") as fout:
+            with open(propose_upgrade, "w") as fout:
```
berland, line 172: rebase mess again ? ;)

komodo/check_up_to_date_pypi.py
```diff
@@ -147,9 +148,14 @@ def run_check_up_to_date(
...
     ),
     propose_upgrade=False,
+    ignore=None,
 ):
     yaml = yaml_parser()
-    releases = load_from_file(yaml, release_file)
+    releases: dict = load_from_file(yaml, release_file)
+    if ignore:
+        for package_name, package_version in list(releases.items()):
+            if re.search(ignore, f"{package_name} {package_version}"):
```
berland, line 157: What if we have some pypi package that includes "main" in its package name? We should probably be explicit whether the regexp refer to the version or name.

tests/test_up_to_date_pypi.py
```diff
@@ -547,3 +547,130 @@ def side_effect(url: str, timeout):
...
+        maintainer: scout
+
+      main:
+        source: pypi
+        make: pip
+        maintainer: scout
+    """
+
+    with tmpdir.as_cwd():
+        with open("repository_file.yml", "w") as fout:
```
berland, line 613: encoding="utf-8"...

berland: Personally I prefer 

```
from pathlib import Path

Path("repository_file.yml").write_text(repository_file_contents, encoding="utf-8")
```

always over  `with open()`

See e.g. https://treyhunner.com/2018/12/why-you-should-be-using-pathlib/

komodo/check_up_to_date_pypi.py
jonathan-eq: Either dict comprehension or dict(filter(filter_function, releases.items()))

tests/test_up_to_date_pypi.py
jonathan-eq: I might start using pathlib for working with paths, but it does not seem like such a big difference for writing or reading a file in the current directory.

komodo/check_up_to_date_pypi.py
```diff
@@ -161,16 +167,20 @@ def run_check_up_to_date(
...
             write_to_file(repository, repository_file)
 
         major_upgrades, minor_upgrades, patch_upgrades, other_upgrades = [], [], [], []
         for name, versions in upgrade_proposals_from_pypi.items():
             pypi_latest = get_version.Version(versions["suggested"])
-            current_version = get_version.Version(versions["previous"])
+            try:
+                current_version = get_version.Version(versions["previous"])
+            except get_version.InvalidVersion:
+                print(f"Could not parse version {versions['previous']}")
```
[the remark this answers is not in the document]
jonathan-eq: Yes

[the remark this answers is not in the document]
berland: It's been too long to remember, but can this create noise in the output? `main` is an accepted version in bleeding.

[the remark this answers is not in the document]
jonathan-eq: Yes, but this only runs on packages which have pypi as a source in repository.yml. Main versions are sourced elsewhere, so it is filtered out at an earlier stage.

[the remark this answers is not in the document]
jonathan-eq: ^usually sourced with git

jonathan-eq:
There is another issue with the code. It skips versions marked as pre-release, but certain packages only have pre-release versions. In this case, it will try to run max() over an empty list and fail.

berland (approved):
Nice improvement!

PR #388, opened 2023-09-05
Update komodo scripts and tests to use yaml file types

### LINKED ISSUE #377: Validate YAML config files as part of linting

jonathan-eq:
Resolves #377 
Resolves https://github.com/equinor/komodo-releases/issues/4240
Resolves https://github.com/equinor/komodo-releases/issues/3492
Resolves https://github.com/equinor/komodo-releases/issues/4381

Blocked by #391 
Depends on #391


komodo/check_up_to_date_pypi.py
```diff
@@ -130,32 +133,81 @@ def insert_upgrade_proposals(upgrade_proposals, repository, releases):
...
+        python_version,
+    )
+    if upgrade_proposals_from_pypi:
+        if propose_upgrade:
+            insert_upgrade_proposals(upgrade_proposals_from_pypi, repository, releases)
+            print(
+                "Writing upgrade proposals from pypi, "
+                "assuming nothing has changed with dependencies..."
+            )
+            with open(propose_upgrade, "w") as fout:
```
berland, line 161: We should always specify encoding when using  `open()`. This line normally gives an error on this, but we have added `unspecified-encoding` to `pyproject.toml` because we haven't gotten this far yet with solving pylint issues in komodo.

komodo/check_up_to_date_pypi.py
```diff
@@ -164,44 +216,28 @@ def main():
...
-            errors.append(
-                f"{name} not at latest pypi version: {pypi_latest}, "
-                f"is at: {current_version}"
-            )
-        sys.exit("\n".join(errors) + "\nFound out of date packages!")
+    print(f"Checking against python version: {args.python_version}")
 
-    else:
-        print("All packages up to date!!!")
+    ReleaseFile()(args.release_file)
```
berland, line 235: These two lines look odd, I suppose the intention is to do the validation. Is it possible to do something such that this line looks less odd and perhaps includes the string "validate" in some way?

komodo/cleanup.py
```diff
@@ -1,35 +1,33 @@
...
-        with open(fname, "r", encoding="utf-8") as f:
-            rels.append(yml.safe_load(f))
 
-    if not isinstance(repo, dict):
-        raise ValueError(f"Malformed package file: {type(repo)}")
-    for rel in rels:
-        if not isinstance(rel, dict):
-            raise ValueError(f"Malformed repository file: {type(rel)}")
+def cleanup(repository_file_path: str, release_files_path: List[str]):
+    repository_file = RepositoryFile()(value=repository_file_path)
```
berland, line 10: "value" looks like an odd argument name here.

komodo/check_up_to_date_pypi.py
jonathan-eq: .Yes, I can put them in a separate function.

komodo/cleanup.py
jonathan-eq: Yes, agreed

komodo/yaml_file_types.py
```diff
@@ -197,21 +197,36 @@ def validate_package_entry(
...
+                raise KomodoException(
+                    f"Package '{package_name}' not found in repository. Did you mean"
+                    f" '{real_packagename}'?"
+                )
             raise KomodoException(f"Package '{package_name}' not found in repository")
         if package_version not in repository_entries[package_name]:
+            if f"v{package_version}" in repository_entries[package_name]:
+                raise KomodoException(
+                    f"Version '{package_version}' of package '{package_name}' not found"
+                    f" in repository. Did you mean 'v{package_version}'?"
```
berland, line 215: That `v` in front of `package_version` should perhaps not be there? I think there are version tags in komodo both including and not including that "v", so for some packages you might get two v's with this.

jonathan-eq: If you want version v1.2.3 but type 1.2.3 instead, this will notify you that the existing version does not exist, but that you might mean v1.2.3 which does exist. We do the same for when the entered package_name's capitalisation does not match the one in repositoryyaml. Example PYTest instead of pytest.

berland: ah, I think I missed the `v` in line 212.

komodo/cli.py
```diff
@@ -60,14 +60,15 @@ def _print_timings(
...
     abs_prefix = Path(args.prefix).resolve()
 
     data = Data(extra_data_dirs=args.extra_data_dirs)
 
     if args.download or (not args.build and not args.install):
         start_time = datetime.datetime.now()
-        git_hashes = fetch(args.pkgs, args.repo, outdir=args.downloads, pip=args.pip)
+        git_hashes = fetch(
+            args.pkgs.content, args.repo.content, outdir=args.downloads, pip=args.pip
+        )
```
berland, line 71: Where does this change stem from? Is it a newer black, or an editor that has done something with the whitespace?

berland (approved):
Phew, large job! well done

Lets merge.

PR #389, opened 2023-09-05
Add linter for upgrade proposals yml files

### LINKED ISSUE #398: Add linter for upgrade_proposal file
Add a linter for linting upgrade_proposal file

jonathan-eq:
Added separate linter for liting upgrade proposals yml files from github actions workflow, before starting building release.
Resolves #398

Depends on #391


komodo/lint_upgrade_proposals.py
```diff
@@ -0,0 +1,63 @@
...
+                    upgrade_proposals_package, upgrade_proposals_package_version
+                )
+                print(
+                    f"Found package: '{upgrade_proposals_package}' with version"
+                    f" {upgrade_proposals_package_version} in repository"
+                )
+            except KomodoException as e:
+                errors.append(e.error)
+        if errors:
+            raise SystemExit(" \n ".join(errors))
```
berland, line 30: whitespace before `\n` should not be needed. Is the next `space` giving an intended visual effect indenting the output?

komodo/lint_upgrade_proposals.py
```diff
@@ -0,0 +1,63 @@
...
+    if releases_with_upgrades_counter == 0:
+        raise ValueError("No upgrades found")
+    if releases_with_upgrades_counter > 1:
+        raise ValueError("Found upgrades for more than one release")
```
berland, line 34: is this an error condition? 🤔

berland: And why mixing `SystemExit` and `ValueError`?

komodo/lint_upgrade_proposals.py
jonathan-eq: I do not remember why I added it, probably for some visual reasons. Removing it now.

komodo/lint_upgrade_proposals.py
jonathan-eq: It is not an error, so it should be systemexit instead of valueerror.

berland (approved):
Looks good!

PR #391, opened 2023-09-05
add-yaml-file-types-for-linting-and-validation

jonathan-eq:


berland:
Split the file rename into two commits, one just doing the filename change, and then a commit that edits it (and the other file edits)

komodo/yaml_file_type.py
```diff
@@ -12,24 +52,80 @@ def __init__(self, *args, **kwargs):
 
     def __call__(self, value):
         file_handle = super().__call__(value)
-        yaml = yml.safe_load(file_handle)
+        yaml = YAML()
+        yml = yaml.load(file_handle)
```
berland, line 56: The variable names look random here since `yaml` and `yml` have very different type. 

Maybe write it as one-line:  `yml  = YAML().load(file_handle)`

komodo/yaml_file_type.py
```diff
@@ -38,16 +134,506 @@ class ManifestFile(YamlFile):
...
         )
-        for _, metadata in yaml.items():
-            assert isinstance(metadata, dict), message
-            assert isinstance(metadata["version"], str), message
-        return yaml
+        assert isinstance(manifest_file_content, dict), message
+        errors = []
+        for package_name, metadata in manifest_file_content.items():
+            if not isinstance(metadata, dict):
+                errors.append(f"Invalid metadata for package '{package_name}'")
```
berland, line 159: yes, `errors.append()` is a better pattern, ref comment above

komodo/yaml_file_type.py
```diff
@@ -38,16 +134,506 @@ class ManifestFile(YamlFile):
...
+
+
+class RepositoryFile(YamlFile):
+    """
+    Return the data from 'repository' YAML, but validate it first.
+    """
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.yml = None
```
berland, line 175: Typing for this class variable would be nice!

komodo/yaml_file_type.py
```diff
@@ -38,16 +134,506 @@ class ManifestFile(YamlFile):
...
+
+    def __call__(self, value: str):
+        self.yml = super().__call__(value)
+        self.validate_repository_file()
+        return self
+
+    def from_yaml_string(self, value: bytes):
+        yaml = YAML()
+        yml = yaml.load(value)
+        self.yml = yml
```
berland, line 185: suggest `self.yml = YAML().load(value)` to avoid unneeded and awkward-to-name temporary variables.

komodo/yaml_file_type.py
```diff
@@ -1,9 +1,49 @@
...
 import os
+from collections import namedtuple
 from pathlib import Path
-from typing import Any, Dict
+from typing import Dict, List
 
-import yaml as yml
+from ruamel.yaml import YAML
+
+komodo_error = namedtuple(
```
berland, line 9: I am a little bit unsure which casing to choose for this. Module constants should be upper-cased, but this is more like a class definition, should we case it as such?

komodo/yaml_file_type.py
berland: I also think that `yml` is a bad variable name, because it is not yaml any longer. Further down you call it `repository_entries`, that might be better.

komodo/yaml_file_type.py
```diff
@@ -38,16 +134,506 @@ class ManifestFile(YamlFile):
...
+    def from_yaml_string(self, value: bytes):
+        yaml = YAML()
+        yml = yaml.load(value)
+        self.yml = yml
+        self.validate_repository_file()
+        return self
+
+    def validate_package_entry(
+        self, package_name: str, package_version: str
+    ) -> komodo_error:
```
berland, line 191: hmm, there is no `return` statement here that indicates it can return a `komodo_error`?

komodo/yaml_file_type.py
```diff
@@ -38,16 +134,506 @@ class ManifestFile(YamlFile):
...
+
+    def lint_maintainer(self, package, version) -> komodo_error:
+        repository_entries = self.yml
+        if package not in repository_entries:
+            raise KomodoException(_komodo_error(package=package, err=MISSING_PACKAGE))
+        if version not in repository_entries[package]:
+            raise KomodoException(
+                _komodo_error(package=package, version=version, err=MISSING_VERSION)
+            )
+        return _komodo_error(
```
berland, line 209: is this an error condition?

komodo/yaml_file_type.py
```diff
@@ -38,16 +134,506 @@ class ManifestFile(YamlFile):
...
+            """setuptools\n      - python\n\npython:\n  "3.8":\n    ..."""
+        )
+        assert isinstance(repository_file_content, dict), message
+        errors = []
+        for package_name, versions in repository_file_content.items():
+            try:
+                Package.validate_package_name(package_name)
+                if not isinstance(versions, dict):
+                    errors.append(
+                        f"Versions of package '{package_name}' is not formatted"
```
berland, line 232: is -> are

komodo/yaml_file_type.py
```diff
@@ -38,16 +134,506 @@ class ManifestFile(YamlFile):
...
+
+    def validate_package_properties(
+        self,
+        package_name: str,
+        package_version: str,
+        package_property: str,
+        package_property_value: str,
+        errors: List[str],
+    ):
+        PRE_CHECKED_PROPERTIES = ["make", "maintainer"]
```
berland, line 276: lower-case, since it is not at module or class level.

komodo/yaml_file_type.py
```diff
@@ -38,16 +134,506 @@ class ManifestFile(YamlFile):
...
+        if package_property in PRE_CHECKED_PROPERTIES:
+            return
+        if package_property == "depends":
+            if not isinstance(package_property_value, list):
+                errors.append(
+                    f"Dependencies for package {package_name} has"
```
berland, line 282: has -> have

komodo/yaml_file_type.py
```diff
@@ -38,16 +134,506 @@ class ManifestFile(YamlFile):
...
+
+    def __call__(self, value: str) -> Dict[str, Dict[str, str]]:
+        yml = super().__call__(value)
+        self.validate_upgrade_proposals_file(yml)
+        self.yml = yml
+        return self
+
+    def from_yaml_string(self, value):
+        yaml = YAML()
+        yml = yaml.load(value)
```
berland, line 327: same comments as above applies here too.

komodo/yaml_file_type.py
```diff
@@ -38,16 +134,506 @@ class ManifestFile(YamlFile):
...
+        self.validate_upgrade_proposals_file(yml)
+        self.yml = yml
+        return self
+
+    def validate_upgrade_key(self, upgrade_key: str) -> None:
+        assert (
```
berland, line 333: Do you want to throw an `AssertionError` here, not something ala `KomodoError`?

komodo/yaml_file_type.py
```diff
@@ -38,16 +134,506 @@ class ManifestFile(YamlFile):
...
+
+    def __call__(self, value: str):
+        yml = super().__call__(value)
+        self.yml = yml
+        self.validate_package_status_file()
+        return self
+
+    def from_yaml_string(self, value: str):
+        yaml = YAML()
+        yml = yaml.load(value)
```
berland, line 388: as above..

berland (changes requested):
halfway through :)

komodo/yaml_file_type.py
jonathan-eq: I changed the key to content, as it refers to the content of the file. Should I call it content_dict instead to better indicate the type? This should be shown in most IDEs as I have added typing, but it could also be more clear.

komodo/yaml_file_type.py
jonathan-eq: Good idea. I did it in separate lines because I was not sure if we had to set options for the yaml loader before using the load method, but it worked with default options.

komodo/yaml_file_type.py
jonathan-eq: This validates that the wanted upgrade_key is in the file, and should exit if it is not present (or invalid?). We could throw KomodoException instead, but I dont think it would make a difference. Other places I use KomodoError if I want to catch the error and extract the message.

komodo/yaml_file_type.py
```diff
@@ -12,24 +52,80 @@ def __init__(self, *args, **kwargs):
...
+            "The file you provided does not appear to be a release file "
+            "produced by komodo. It may be a repository file. Release files "
+            "have a format like the following:\n\n"
+            'python: 3.8.6-builtin\nsetuptools: 68.0.0\nwheel: 0.40.0\nzopfli: "0.3"'
+        )
+        assert isinstance(release_file_content, dict), message
+        errors = []
+        for package_name, package_version in release_file_content.items():
+            Package.validate_package_entry_with_errors(
+                package_name, package_version, errors
```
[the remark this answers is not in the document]
jonathan-eq: Yes, I discussed this with some team members and we agreed on a pattern of always returning lists (empty list for no errors) and extending the initial errors list with the new errors.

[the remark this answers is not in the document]
berland: To return a list of errors is fine (containing zero or more elements), but are you really sure you want to supply a handle to a list to a function and let it modify it? Such a thing is usually an anti-pattern in Python.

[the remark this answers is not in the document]
berland: Ah, it seems you have done this already :)

komodo/yaml_file_types.py
```diff
@@ -12,24 +52,79 @@ def __init__(self, *args, **kwargs):
...
+        errors = []
+        for package_name, package_version in release_file_content.items():
+            error: List[str] = Package.validate_package_entry_with_errors(
+                package_name, package_version
+            )
+            errors.extend(error)
+        handle_validation_errors(errors, message)
+
+    @staticmethod
+    def lint_release_name(packagefile_path: str):
```
berland, line 99: The return type should be typed here.

komodo/yaml_file_types.py
```diff
@@ -38,16 +133,521 @@ class ManifestFile(YamlFile):
...
+        return errors
+
+    def validate_package_properties(
+        self,
+        package_name: str,
+        package_version: str,
+        package_property: str,
+        package_property_value: str,
+    ) -> List[str]:
+        pre_checked_properties = ["make", "maintainer"]
```
berland, line 276: Maybe a docstring is warranted here stating that what is returned is a list of errors.

berland: (and similarly for other new functions). It is not readable from the typing or the function name whether exceptions are thrown or just a list of errors are returned. When neither of these explain this, there should be a docstring stating it.

komodo/yaml_file_types.py
```diff
@@ -38,16 +133,521 @@ class ManifestFile(YamlFile):
...
+        raise TypeError(
+            f"Package '{package_name}' has invalid visibility type"
+            f" ({package_visibility})"
+        )
+
+    @staticmethod
+    def validate_package_maturity(package_name: str, package_maturity: str) -> bool:
+        if isinstance(package_maturity, str):
+            if package_maturity in Package.VALID_MATURITIES:
+                return True
```
berland, line 514: It seems this function is only used in the function `validate_package_maturity_with_errors()`, and then that nothing ever relies on the truthiness of this function. Should this function just be inlined in `validate_package_maturity_with_errors()`?

komodo/yaml_file_types.py
```diff
@@ -38,16 +133,521 @@ class ManifestFile(YamlFile):
...
+            )
+
+
+def handle_validation_errors(errors: List[str], message: str):
+    if errors:
+        raise SystemExit("\n".join(errors + [message]))
+
+
+def load_package_status_file(package_status_string: str):
+    package_status = PackageStatusFile().from_yaml_string(package_status_string)
```
berland, line 647: You can merge these two lines and avoid the declaration and naming of the temporary variable `package_status`

komodo/yaml_file_types.py
```diff
@@ -38,16 +133,521 @@ class ManifestFile(YamlFile):
...
+    return package_status
+
+
+def load_repository_file(repository_file_string):
+    repository_file = RepositoryFile().from_yaml_string(repository_file_string)
+    return repository_file
```
berland, line 653: ditto

berland:
Are the tests for this PR inside PR #388? The tests related to the changes in this PR should be cherry-picked into this PR.

komodo/yaml_file_types.py
jonathan-eq: Agreed!

komodo/yaml_file_types.py
jonathan-eq: What is meant by truthiness of the function? I tried separating the responsibilities into two different functions, one checks the type and the value, and the other one collects the error(s) in the wanted format. It leads to extra code, but could be more reusable. Is it worth the trade off?

komodo/yaml_file_types.py
jonathan-eq: Yes, I agree!

komodo/yaml_file_types.py
berland: I mean that I don't see any statements like `if validate_package_visibility(...):` which is a statement depending on whether that function returns True or False ("truthiness..")

komodo/yaml_file_types.py
```diff
@@ -38,16 +133,544 @@ class ManifestFile(YamlFile):
...
+        """
+        errors = []
+        try:
+            Package.validate_package_importance(package_name, package_importance)
+        except (ValueError, TypeError) as e:
+            errors.append(str(e))
+        return errors
+
+    @staticmethod
+    def validate_package_visibility(package_name: str, package_visibility: str) -> None:
```
berland, line 510: (here the typing is wrong, it says None, but there is a `return True` statement)

tests/test_yaml_file_types.py
```diff
@@ -0,0 +1,930 @@
...
+
+from komodo.yaml_file_types import (
+    KomodoException,
+    PackageStatusFile,
+    ReleaseFile,
+    RepositoryFile,
+    _komodo_error,
+)
+
+VALID_PYTHON_PACKAGE_KV_LIST = 'zopfli: "0.3"\npytest: 0.40.1'
```
berland, line 14: I can't find repeated usage of these constants, I think I would just inline these strings in the tests below. Keeping them as variables could serve a documentation purpose, but it seems that is already covered by usage of the `id` argument to `pytest.param`.

tests/test_yaml_file_types.py
```diff
@@ -0,0 +1,930 @@
...
+            INVALID_YAML,
+            pytest.raises(
+                AssertionError,
+                match=r"The file you provided does not appear to be a release file",
+            ),
+            id="invalid_release_file_format",
+        ),
+        pytest.param(
+            INVALID_PYTHON_PACKAGE_KV_LIST_DUPLICATE_KEYS,
+            pytest.raises(DuplicateKeyError, match='found duplicate key "zopfli"'),
```
berland, line 45: Why do we have a regular exception only here, but all others raise a `SystemExit`?

tests/test_yaml_file_types.py
```diff
@@ -0,0 +1,930 @@
...
+    package_name, package_version, expectation, return_object
+):
+    repo_file = RepositoryFile().from_yaml_string(LINT_MAINTAINER_TEST_REPO)
+    with expectation:
+        result = repo_file.lint_maintainer(package_name, package_version)
+    if return_object:
+        assert result == return_object
+
+
+VALID_PACKAGE_STATUS = (
```
berland, line 846: Strings can probably be inlined in pytest.param.

komodo/yaml_file_types.py
jonathan-eq: Ahhh, okay, I get it now. The return value is never used, so I will change it to return None instead.

tests/test_yaml_file_types.py
jonathan-eq: DuplicateKeyError is a ruamel.yaml exception raised when the yaml parser fails due to duplicate keys. That is the error which was thrown at a later stage in the release process, and which I want linted away early on.

tests/test_yaml_file_types.py
jonathan-eq: This was to avoid having magic-strings in pytest.param, but it should probably be fine as all cases have descriptive ids. I also think some yaml-strings (VALID_REPOSITORY) are reused in some cases. Should I keep the reused ones as variables, and do all single-use ones inline?

jonathan-eq: Although, I will keep the multiline yaml strings (e.g repository dict) as variables as they ruin readability.

tests/test_yaml_file_types.py
berland: But should we try to capture the DuplicateKeyError and reraise it as a SystemExit for consistency?

jonathan-eq: Yes, we can absolutely do that

berland (approved):
Thanks for the perserverence, lets ship 🚢

PR #393, opened 2023-09-19
Symlink create links should tell when stable implicitly moved

### LINKED ISSUE #392: create_links should tell when it is the stable link that is moved
We are getting messages from create_links like this:

> Existing symlink 2023.06-py38 moved from 2023.06.10-py38 to 2023.06.11-py38

which are nice, but it would be even nicer if it also could state that "stable" is actually pointing to this, and thus stable is impliclitly moved.

jonathan-eq:
Resolves #392 
Depends on #391

### DIFF
--- a/komodo/symlink/create_links.py
+++ b/komodo/symlink/create_links.py
@@ -3,6 +3,7 @@ import json
 import os
 import sys
 from contextlib import contextmanager
+from typing import List
 
 from .sanity_check import verify_integrity
 
@@ -15,6 +16,19 @@ def working_dir(path):
         yield
     finally:
         os.chdir(prev_dir)
+
+
+def get_implicitly_moved_symlinks(key: str, link_dict: dict) -> List[str]:
+    sources = [src for src, dst in link_dict.items() if dst == key]
+    implicitly_moved_symlinks = []
+    for source in sources:
+        implicitly_moved_symlinks.extend(
+            get_implicitly_moved_symlinks(source, link_dict)
+        )
+
+    if not sources:
+        implicitly_moved_symlinks.append(key)
+    return implicitly_moved_symlinks
 
 
 def _create_link(src, dst, link_dict):
@@ -30,7 +44,19 @@ def _create_link(src, dst, link_dict):
         if existing_link == src:
             return
         os.remove(dst)
-        print(f"Existing symlink {dst} moved from {existing_link} to {src}")
+        implicitly_moved_sources = get_implicitly_moved_symlinks(dst, link_dict)
+        if dst in implicitly_moved_sources:
+            implicitly_moved_sources.remove(dst)
+
+        message = f"Existing symlink {dst} moved from {existing_link} to {src}." + (
+            (
+                " Some symlinks were implicitly moved due to this: "
+                f"{', '.join(implicitly_moved_sources)}."
+            )
+            if implicitly_moved_sources
+            else ""
+        )
+        print(message)
     else:
         print(f"Created new symlink {dst} pointing to {src}")
 
--- a/tests/test_link_io_structure.py
+++ b/tests/test_link_io_structure.py
@@ -95,7 +95,36 @@ def test_overwrite_symlink_stdout(tmpdir, capsys):
         os.symlink("2023.06", "2023")
         create_symlinks(links_dict)
         captured = capsys.readouterr()
-        assert captured.out == "Existing symlink 2023 moved from 2023.06 to 2023.07\n"
+        print(captured.out)
+        assert captured.out == "Existing symlink 2023 moved from 2023.06 to 2023.07.\n"
+
+
+def test_overwrite_symlink_implicit_multiple_stdout(tmpdir, capsys):
+    links_dict = {
+        "root_folder": tmpdir,
+        "links": {
+            "stable": "stable-py38",
+            "stable-py38": "2023.09.03",
+            "azure-stable": "azure-stable-py38",
+            "azure-stable-py38": "stable-py38",
+            "2023.09.03": "2023.09.03-rc4",
+        },
+    }
+    with tmpdir.as_cwd():
+        os.mkdir("2023.09.03-rc4")
+        os.mkdir("2023.09.03-rc3")
+        os.symlink("stable-py38", "stable")
+        os.symlink("azure-stable-py38", "azure-stable")
+        os.symlink("stable-py38", "azure-stable-py38")
+        os.symlink("2023.09.03", "stable-py38")
+        os.symlink("2023.09.03-rc3", "2023.09.03")
+        create_symlinks(links_dict)
+        captured = capsys.readouterr()
+        print(captured.out)
+        assert (
+            captured.out
+            == "Existing symlink 2023.09.03 moved from 2023.09.03-rc3 to 2023.09.03-rc4. Some symlinks were implicitly moved due to this: stable, azure-stable.\n"
+        )
 
 
 def test_integration(tmpdir):


komodo/symlink/create_links.py
```diff
@@ -17,6 +17,13 @@ def working_dir(path):
         os.chdir(prev_dir)
 
 
+def traverse_symlink_tree(key, link_dict):
+    values = [k for k, v in link_dict.items() if v == key]
+    for value in values:
+        traverse_symlink_tree(value, link_dict)
+        print(f"{value} was implicitly moved as a side effect")
```
berland, line 24: I think we can drop "as a side effect". This is the main effect that we target after all.

jonathan-eq: Yes. sounds good!

komodo/symlink/create_links.py
```diff
@@ -43,10 +51,10 @@ def create_symlinks(links_dict):
 
     if not os.path.isdir(root_folder):
         raise ValueError(f"{root_folder} is not a directory or does not exist")
-
+    links_dict_links = links_dict["links"]
     with working_dir(root_folder):
         for dst, src in links_dict["links"].items():
-            _create_link(src, dst, links_dict["links"])
+            _create_link(src, dst, links_dict_links)
```
berland, line 57: why this change?

komodo/symlink/create_links.py
```diff
@@ -17,6 +17,13 @@ def working_dir(path):
         os.chdir(prev_dir)
 
 
+def traverse_symlink_tree(key, link_dict):
```
berland, line 20: This should be renamed to something that describes what it is doing (locating implicitly moved symlinks), not how it is doing it.

komodo/symlink/create_links.py
```diff
@@ -17,6 +17,13 @@ def working_dir(path):
...
+    values = [k for k, v in link_dict.items() if v == key]
```
berland, line 21: (pylint is not fond of single-letter variables..)

tests/test_link_io_structure.py
```diff
@@ -95,7 +95,32 @@ def test_overwrite_symlink_stdout(tmpdir, capsys):
         os.symlink("2023.06", "2023")
         create_symlinks(links_dict)
         captured = capsys.readouterr()
-        assert "Existing symlink 2023 moved from 2023.06 to 2023.07\n" == captured.out
+        print(captured.out)
```
berland, line 98: leftover print?

komodo/symlink/create_links.py
```diff
@@ -17,6 +17,13 @@ def working_dir(path):
...
+    for value in values:
+        traverse_symlink_tree(value, link_dict)
+        print(f"{value} was implicitly moved")
```
berland, line 24: This means we will get one line of output for every level in the link-tree. Will that lead to too much noise in the slack message? Perhaps we only want the topmost change, or it might be possible to give all the information in one line.

komodo/symlink/create_links.py
jonathan-eq: I used the value multiple places in previous attempt, so I had it as a variable instead. Removing the change now.

komodo/symlink/create_links.py
jonathan-eq: Yes, agreed

tests/test_link_io_structure.py
jonathan-eq: Yes, is it possible to configure pylint to catch leftover print statements in tests?

komodo/symlink/create_links.py
jonathan-eq: I don't know how long a symlink-chain usually is, but it would be better to know too much than not enough. I will try out different formattings, maybe we can shorten it without leaving out information.

tests/test_link_io_structure.py
berland: Don't think so.

komodo/symlink/create_links.py
```diff
@@ -17,6 +17,15 @@ def working_dir(path):
         os.chdir(prev_dir)
 
 
+def locate_implicitly_moved_symlinks(key, link_dict):
```
berland, line 20: type this function definition?

komodo/symlink/create_links.py
```diff
@@ -30,6 +39,7 @@ def _create_link(src, dst, link_dict):
             return
         os.remove(dst)
         print(f"Existing symlink {dst} moved from {existing_link} to {src}")
+        locate_implicitly_moved_symlinks(dst, link_dict)
```
berland, line 42: It is hard to read from the function name that this has a side-effect to do a `print`. I think it should return a string that we can inject into the print statement on the line above (then we can format it inside a paranthesis e.g).

komodo/symlink/create_links.py
```diff
@@ -30,6 +44,9 @@ def _create_link(src, dst, link_dict):
             return
         os.remove(dst)
         print(f"Existing symlink {dst} moved from {existing_link} to {src}")
+        implicitly_moved_sources = get_implicitly_moved_symlinks(dst, link_dict)
+        for source in implicitly_moved_sources:
+            print(f"{source} was implicitly moved")
```
berland, line 49: I think I only want the last one, would that the first or last element of the list? And since this is to be sent to a Slack message, I would like to have it formatted as a sentence to read. What happens with the lineshifts from `print` statements is a little hard to predict.

tests/test_link_io_structure.py
```diff
@@ -95,7 +95,30 @@ def test_overwrite_symlink_stdout(tmpdir, capsys):
         os.symlink("2023.06", "2023")
         create_symlinks(links_dict)
         captured = capsys.readouterr()
-        assert "Existing symlink 2023 moved from 2023.06 to 2023.07\n" == captured.out
+        assert (
+            "Existing symlink 2023 moved from 2023.06 to 2023.07"
+            in captured.out.split("\n")
```
berland, line 100: This assert now leaves the possibility for garbage on subsequent lines. Can you constrain it to not allow that?

tests/test_link_io_structure.py
```diff
@@ -95,7 +95,30 @@ def test_overwrite_symlink_stdout(tmpdir, capsys):
...
+        os.mkdir("g")
+        os.symlink("d", "a")
+        os.symlink("c", "b")
+        os.symlink("d", "c")
+        os.symlink("e", "d")
+        os.symlink("f", "e")
+        create_symlinks(links_dict)
+        captured = capsys.readouterr()
+        for symlink in list("ab"):
+            assert f"{symlink} was implicitly moved" in captured.out.split("\n")
```
berland, line 120: This example with single letters is fine, but it is hard to read and verify for correctness. Can you make another example that is using `stable` -> `stable-py38` -> `2023.09.03` f.ex and assert exactly on the full output?

berland (changes requested):

tests/test_link_io_structure.py
jonathan-eq: Yes, certainly. It was easier to read when I had the drawing next to me on my desk

komodo/symlink/create_links.py
jonathan-eq: Yes, the method should just return the name of the symlink that was moved, so we can format it later on.

komodo/symlink/create_links.py
jonathan-eq: I updated it. Can you verify if it looks good now / how you would improve it?

komodo/symlink/create_links.py
```diff
@@ -30,7 +44,15 @@ def _create_link(src, dst, link_dict):
         if existing_link == src:
             return
         os.remove(dst)
-        print(f"Existing symlink {dst} moved from {existing_link} to {src}")
+        implicitly_moved_sources = get_implicitly_moved_symlinks(dst, link_dict)
+        message = (
+            f"Existing symlink {dst} moved from {existing_link} to {src}."
+            + " Some symlinks were implicitly moved due to this."
```
berland, line 50: Looking better! I would move the punctuation dot to after the paranthesis. Or now that it is more explicit how it will look in Slack, maybe a colon and then just a comma-separated list without paranthesis.

jonathan-eq:
I found an edge case for when there was only one symlink (no chain), then the symlink move would be mentioned twice.
Added some code to check if this is the case, then pop it from the returned list if it is there.

berland (approved):
👌🏻

PR #394, opened 2023-09-20
Add platform in symlink proposal PR

jonathan-eq:
Added reference to which platform the symlink will be updated for in both the PR title and body.
Resolves [#3869](https://github.com/equinor/komodo-releases/issues/3869)


xjules (approved):
Very good! :rocket:

PR #401, opened 2023-09-27
Be explicit about stacklevel

berland:
This ensure that the context is preserved when we issue warnings using function wrappers.

Solves flake B028, which for some reason does not seem to trigger on Python 3.10, but does so locally on Python 3.8


jonathan-eq (approved):
LGTM!!

PR #404, opened 2023-09-28
Pin urllib3 while testing

berland:
This is for compatibility with RHEL7 nodes


jonathan-eq (approved):
LGTM!

berland:
This one sneaked its way in in #405

PR #408, opened 2023-10-02
Fix ReleaseDir class to load *yml files

berland:
This fixes two bugs:
* komodo-transpiler writes yaml files with yml-suffix, while this function assumed it was yaml as suffix.
* ReleaseFile()(yaml_file) did not give a generator that  can use, changed to use content and put it in a subdictionary.


jonathan-eq (approved):
Good job!

PR #411, opened 2023-10-13
Allow deleting symlinks as well, not only directories

berland:


berland:
Context: https://github.com/equinor/komodo-releases/actions/runs/6502849750/job/17662505292

eivindjahren (approved):

PR #412, opened 2023-10-13
Utilize pathlib in test_switch.py

berland:
(drive-by improvement)

### DIFF
--- a/tests/test_switch.py
+++ b/tests/test_switch.py
@@ -1,5 +1,3 @@
-import os
-
 import pytest
 
 from komodo import switch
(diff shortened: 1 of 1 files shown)


tests/test_switch.py
```diff
@@ -26,16 +25,16 @@ def test_write_activator_switches(tmpdir):
 else
     echo -e "{MIGRATION_WARNING}"
 fi
-"""
+""".strip()
+    )
 
-        assert actual.read() == expected
+    # assert actual.read() == expected
```
kwinkunks, line 31: Did you mean to leave this here?

berland: no, thanks!

kwinkunks (approved):
Other than the possibly extraneous comment, this looks great. So much cleaner than os.path, especially as tmpdir arrives as a Path. 👍

tests/test_switch.py
berland: fixed

PR #415, opened 2023-10-30
Fix komodo lint maturity not ignoring file extension

jonathan-eq:
The komodo lint maturity script fails when testing locally because it is not removing the file extension before checking if the release name is valid. The commit also updates the outdated tests.


berland (approved):
LGTM

PR #417, opened 2023-11-01
Allow a rm command to fail

berland:
The rm <build>.delete statement has been observed to be flaky on NFS filesystems, probably due to leftover .nfs* files blocking deletion of the directories.

We fix this by moving the old release to a unique filename, and let the rm command fail (will be logged to stdout). Subsequent builds will do the cleanup.


jonathan-eq (approved):
LGTM!

PR #418, opened 2023-11-02
Fix upgrade-proposals linter bug for multiple releases

jonathan-eq:
The new upgrade-proposals linter threw an error if there were multiple releases being upgraded at the same time. This is an intended feature, so that is being removed from the linter in this commit.


xjules (approved):
Nice job @jonathan-eq !

PR #426, opened 2024-01-03
Remove statoil proxy from build pyside-dev script

### LINKED ISSUE #420: Remove proxy before mid-January 2024
We are setting the proxy for a download 

https://github.com/equinor/komodo/blob/c33a64eb368354c14714e7346777dbb1ec55f1d0/komodo/data/build__pyside-dev.sh#L131

Once qt.io is approved in the firewall, we can remove the proxy setting.

jonathan-eq:
Resolves #420


kwinkunks (approved):
Perfect

PR #430, opened 2024-01-12
Pylint improvement fix unused-argument

jonathan-eq:


komodo/build.py
```diff
@@ -42,7 +42,7 @@ def dfs(pkg, version, pkgs, repo):
     return flatten(dependencies)
 
 
-def rpm(pkg, ver, path, data, prefix, *args, **kwargs):
+def rpm(pkg, ver, path, prefix):
```
kwinkunks, line 45: Big chage, do you know for sure nothing is calling this function with positional args, or passing `*args` or `**kwargs`?

jonathan-eq: It should only be called from the make() function, but that is an assumption

kwinkunks: After chatting, I think I get it.

komodo/build.py
```diff
@@ -64,20 +64,19 @@ def cmake(
     pkg,
     ver,
     path,
```
kwinkunks, line 66: Use `pkgpath` here to match other functions?

komodo/build.py
kwinkunks, line 45: Use `pkgpath` here to match other functions?

komodo/build.py
```diff
@@ -42,7 +42,7 @@ def dfs(pkg, version, pkgs, repo):
     return flatten(dependencies)
 
 
-def rpm(pkg, ver, path, data, prefix, *args, **kwargs):
+def rpm(pkg, ver, path, prefix):
     # cpio always outputs to cwd, can't be overriden with switches
```
kwinkunks, line 46: Recommend adding docstring to this and all other functions. Might save the next person a bit of time figuring this module out.

komodo/build.py
```diff
@@ -313,30 +317,75 @@ def resolve(x):
...
-            hash=current.get("hash"),
-        )
+            makeopts = f"{makeopts} {extra_makeopts}"
+        makeopts = resolve(makeopts)
+
+        if make == "rpm":
+            rpm(pkg=package_name, ver=ver, path=pkgpath, prefix=prefix)
+        if make == "cmake":
+            cmake(
+                pkg=package_name,
```
kwinkunks, line 329: These could match too...

komodo/build.py
jonathan-eq: I think that is a separate pylint issue, but it is definitely on the to-do list!

kwinkunks (approved):
Looks good Jonathan :)

Issue #436, opened 2024-01-19
komodo-show-version does not work with python3.11 and komodoenv


larsevj:
The behaviour of venv has changed from python3.8 -> 3.11, and this is causing komodo-show-version to fail.

PR #437, opened 2024-01-19
Fixes komodo-show-version

larsevj:
Resolves: #436 
This seems to work for now, but might be a more robust or better way to handle this.

### DIFF
--- a/komodo/data/build__python-virtualenv.sh
+++ b/komodo/data/build__python-virtualenv.sh
@@ -26,4 +26,5 @@ while test $# -gt 0; do
     esac
     shift
 done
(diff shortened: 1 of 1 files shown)


komodo/data/build__python-virtualenv.sh
```diff
@@ -26,4 +26,5 @@ while test $# -gt 0; do
     esac
     shift
 done
-$VIRTUALENV -p $VIRTUALENV_INTERPRETER -v --always-copy --never-download --no-setuptools --no-pip --no-wheel ${FAKEROOT}/${PREFIX} 1>&2
+# $VIRTUALENV -p $VIRTUALENV_INTERPRETER -v --always-copy --never-download --no-setuptools --no-pip --no-wheel ${FAKEROOT}/${PREFIX} 1>&2
```
kwinkunks, line 29: Suggest deleting this, or commenting why we need to keep it.

PR #439, opened 2024-01-22
Use globbing prior to issuing rm command

### LINKED ISSUE #433: shell() does not expand bash wildcard
The wildcard `*` is not expanded in these lines causing all nightly bleeding builds to be left on disk:

https://github.com/equinor/komodo/blob/6e99cc528b1a4ec2d01ee8c113822af91f85acad/komodo/cli.py#L166-L170

andreas-el:
Resolves #433
Resolves https://github.com/equinor/komodo-releases/issues/4983

### DIFF
--- a/komodo/cli.py
+++ b/komodo/cli.py
@@ -1,6 +1,7 @@
 import argparse
 import contextlib
 import datetime
+import glob
 import os
 import sys
 import uuid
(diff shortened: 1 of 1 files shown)


kwinkunks (approved):
Looks great, let's do it.

PR #440, opened 2024-01-22
Add guard to avoid overwriting releases

berland:
Only bleeding builds will be overwritten by default.

Solves the komodo part of https://github.com/equinor/komodo-releases/issues/5063

### DIFF
--- a/komodo/cli.py
+++ b/komodo/cli.py
@@ -82,6 +82,9 @@ def _main(args):
     if args.download and not args.build:
         sys.exit(0)
 
+    if (Path(args.prefix) / args.release).exists() and "bleeding" not in args.release:
+        raise RuntimeError("Only bleeding builds can be overwritten")
+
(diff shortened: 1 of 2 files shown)


komodo/cli.py
```diff
@@ -82,6 +82,9 @@ def _main(args):
     if args.download and not args.build:
         sys.exit(0)
 
+    if (Path(args.prefix) / args.release).exists() and "bleeding" not in args.release:
```
xjules, line 85: I understood that we want to have an option that can still overwrite the release in the action that is.

berland: right!

berland: fixed.

berland:
See also https://github.com/equinor/komodo-releases/pull/5076

jonathan-eq (approved):
Great improvement!

PR #441, opened 2024-01-22
Fix activator switch

### LINKED ISSUE #438: komodo enable script is not compatible with rhel8
With the current activator script, if you try to source any release not using the full release name (including rhel7 or rhel8) you will get a migration warning if you are using rhel8 saying that rhel6 is not supported:
``` 
Attention! Your machine is running on an environment that is not supported. RHEL6 has been phased out.
From October 2020, komodo versions only support RHEL7.
```
This because we only check for rhel7 in the enable script. We need to make this compatible with two rhel versions and two python versions. Where the initial plan is to provide py38 release for rhel7+8 and py311 release for rhel8.

larsevj:
Resolves #438

### DIFF
--- a/komodo/data/activator_switch.csh.tmpl
+++ b/komodo/data/activator_switch.csh.tmpl
@@ -1,3 +1,4 @@
+{% if py_version == "py38" -%}
 if ( `uname -r` =~ *el7* ) then
     setenv KOMODO_ROOT {{ prefix }}
     set KOMODO_RELEASE_REAL = "{{ release }}"
@@ -7,6 +8,30 @@ if ( `uname -r` =~ *el7* ) then
         set prompt = "[$KOMODO_RELEASE_REAL] $_KOMODO_OLD_PROMPT"
     endif
     setenv KOMODO_RELEASE $KOMODO_RELEASE_REAL
-else
+elif ( `uname -r` =~ *el8* ) then
+    setenv KOMODO_ROOT {{ prefix }}
+    set KOMODO_RELEASE_REAL = "{{ release }}"
+
+    source $KOMODO_ROOT/$KOMODO_RELEASE_REAL-rhel8/enable.csh
+    if ( $?_KOMODO_OLD_PROMPT ) then
+        set prompt = "[$KOMODO_RELEASE_REAL] $_KOMODO_OLD_PROMPT"
+    endif
+    setenv KOMODO_RELEASE $KOMODO_RELEASE_REAL
+{%- else -%}
+if ( `uname -r` =~ *el7* ) then
+    echo "Error! Komodo release for Python newer than 3.8 is not available on RHEL7."
+elif ( `uname -r` =~ *el8* ) then
+    setenv KOMODO_ROOT {{ prefix }}
+    set KOMODO_RELEASE_REAL = "{{ release }}"
+
+    source $KOMODO_ROOT/$KOMODO_RELEASE_REAL-rhel8/enable.csh
+    if ( $?_KOMODO_OLD_PROMPT ) then
+        set prompt = "[$KOMODO_RELEASE_REAL] $_KOMODO_OLD_PROMPT"
+    endif
+    setenv KOMODO_RELEASE $KOMODO_RELEASE_REAL
+{%- endif %}
+elif ( `uname -r` =~ *el6* ) then
     echo -e "{{ migration_warning }}"
+else
+    echo "Attention! Your machine is running on an environment that is not supported."
 endif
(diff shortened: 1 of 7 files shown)


tests/test_switch.py
```diff
@@ -7,23 +7,33 @@
 
 def test_write_activator_switches(tmpdir):
     prefix = tmpdir / "prefix"
-    release = "2020.01.01-py27-rhel6"
-    expected_release = "2020.01.01-py27"
+    release = "2020.01.01-py38-rhel6"
+    expected_release = "2020.01.01-py38"
```
andreas-el, line 11: What's the reason for this update? Did we have py38 on rhel6 even?

komodo/data/activator_switch.csh.tmpl
```diff
@@ -7,6 +8,31 @@ if ( `uname -r` =~ *el7* ) then
...
+    if ( $?_KOMODO_OLD_PROMPT ) then
+        set prompt = "[$KOMODO_RELEASE_REAL] $_KOMODO_OLD_PROMPT"
+    endif
+    setenv KOMODO_RELEASE $KOMODO_RELEASE_REAL
+{%- endif %}
+elif ( `uname -r` =~ *el6* ) then
     echo -e "{{ migration_warning }}"
+else
+    echo -e "Attention! Your machine is running on an environment "
+    "that is not supported."
```
berland, line 37: missing echo on line 37

komodo/data/activator_switch.csh.tmpl
```diff
@@ -7,6 +8,31 @@ if ( `uname -r` =~ *el7* ) then
         set prompt = "[$KOMODO_RELEASE_REAL] $_KOMODO_OLD_PROMPT"
     endif
     setenv KOMODO_RELEASE $KOMODO_RELEASE_REAL
-else
+elif ( `uname -r` =~ *el8* ) then
```
berland, line 11: backticks in shell is deprecated, should use `$( ... )`

komodo/data/activator_switch.tmpl
```diff
@@ -1,10 +1,32 @@
...
+
+    source $KOMODO_ROOT/$KOMODO_RELEASE_REAL-rhel8/enable
+    export PS1="(${KOMODO_RELEASE_REAL}) ${_PRE_KOMODO_PS1}"
+    export KOMODO_RELEASE=$KOMODO_RELEASE_REAL
+{%- endif %}
+elif [[ $(uname -r) == *el6* ]]; then
     echo -e "{{ migration_warning }}"
+else
+    echo -e "Attention! Your machine is running on an environment "
+    "that is not supported."
```
berland, line 31: missing echo

komodo/data/activator_switch.tmpl
```diff
@@ -1,10 +1,32 @@
...
+elif [[ $(uname -r) == *el8* ]] ; then
+    export KOMODO_ROOT={{ prefix }}
+    KOMODO_RELEASE_REAL={{ release }}
+
+    source $KOMODO_ROOT/$KOMODO_RELEASE_REAL-rhel8/enable
+    export PS1="(${KOMODO_RELEASE_REAL}) ${_PRE_KOMODO_PS1}"
+    export KOMODO_RELEASE=$KOMODO_RELEASE_REAL
+{%- else -%}
+if [[ $(uname -r) == *el7* ]] ; then
+    echo -e "Warning! Komodo release for python3.11 is not available on RHEL7."
```
berland, line 18: should we write for "python newer than 3.8" to be in sync with what the tests actually tests?

tests/test_switch.py
```diff
@@ -41,13 +51,44 @@ def test_write_activator_switches(tmpdir):
...
+    source $KOMODO_ROOT/$KOMODO_RELEASE_REAL-rhel8/enable.csh
+    if ( $?_KOMODO_OLD_PROMPT ) then
+        set prompt = "[$KOMODO_RELEASE_REAL] $_KOMODO_OLD_PROMPT"
+    endif
+    setenv KOMODO_RELEASE $KOMODO_RELEASE_REAL
+elif ( `uname -r` =~ *el6* ) then
     echo -e "{MIGRATION_WARNING}"
+else
+    echo -e "Attention! Your machine is running on an environment "
+    "that is not supported."
```
berland, line 67: echo

tests/test_switch.py
andreas-el: I guess this is okay as is. I guess i'm just a bit put off by the combination in the version number. If the test only verifies that the suffixes are picked correctly, we could just as well update the rest of the version string to something more recent to avoid confusion?

komodo/data/activator_switch.tmpl
berland: Yup, this is an error, not warning.

andreas-el (approved):
LGTM 🚀

PR #442, opened 2024-01-23
Use current release to find package version from komodoenv

### LINKED ISSUE #436: komodo-show-version does not work with python3.11 and komodoenv

larsevj:
Resolves #436


kwinkunks (approved):
This looks good to me.

PR #449, opened 2024-01-26
Use echo on each line in migration message for activator switch in csh

larsevj:


tests/test_switch.py
```diff
@@ -60,7 +60,9 @@ def test_write_activator_switches(tmpdir):
     endif
     setenv KOMODO_RELEASE $KOMODO_RELEASE_REAL
 else if ( `uname -r` =~ *el6* ) then
-    echo -e "{MIGRATION_WARNING}"
+    echo "Attention! Your machine is running on an environment that is not supported. RHEL6 has been phased out."
```
berland, line 63: why do you want to inline `komodo.switch.MIGRATION_WARNING` ?

PR #450, opened 2024-01-29
Improve check-pypi script

### LINKED ISSUE #446: Consider adding a flag to `komodo-check-pypi` to check only minor/patch version changes
Currently the script will suggest the upgrade to the newest available tag, while it could be nice to be able to choose whether you need major upgrades or not

### LINKED ISSUE #447: check-pypi script showing os incompatible upgrades
check-pypi for packages might suggest packages that are incompatible with the target OS / platform. An example of this is PyQt5-Qt5. The current latest version is 5.15.12. We are currently using 5.15.2, so naturally it suggests upgrading this. Upon building the new release on a linux machine, it failed because it could not find the package. Turns out, 5.15.12 only has wheels for macosx.

check-pypi script should take the platform or OS in as an argument, or default on the OS/platform of the host running the script, and filter out the upgrades not compatible.

jonathan-eq:
Resolves #447
Resolves #446

### DIFF
--- a/komodo/check_up_to_date_pypi.py
+++ b/komodo/check_up_to_date_pypi.py
@@ -71,7 +71,9 @@ def is_platform_compatible(build_info: List[Dict], platform: str) -> bool:
     return False
 
 
-def compatible_versions(releases: dict, python_version, platform: str):
+def compatible_versions(
+    releases: dict, python_version, platform: str
+) -> List[get_version.Version]:
     compatible_versions = []
     for version_str, build_info in releases.items():
         try:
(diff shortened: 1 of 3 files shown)


komodo/check_up_to_date_pypi.py
```diff
@@ -281,6 +292,15 @@ def get_args() -> argparse.Namespace:
             "If given, will ignore packages with names that match the regex expression"
         ),
     )
+    parser.add_argument(
+        "--target-platform",
+        default=sys.platform,
+        choices=["darwin", "win32", "linux"],
+        help=(
+            "Which OS the new package version should be compatible with. Should be either linux,"
+            " macos, or win32. Defaults to OS of host machine"
```
kvashchuka, line 329: small thing - the choices do not match the help

komodo/check_up_to_date_pypi.py
```diff
@@ -56,7 +57,19 @@ def get_python_requirement(sources: list):
     return ""
 
 
-def compatible_versions(releases: dict, python_version):
+def is_platform_compatible(build_info: List[Dict], platform: str) -> bool:
+    if platform.lower() == "darwin":
+        platform = "macos"
```
kvashchuka, line 62: Why do we need this `if`? Could you explain, I do not follow

jonathan-eq: sys.platform on mac returns darwin, but it should be macos in the filename, so I convert it to macos

komodo/check_up_to_date_pypi.py
jonathan-eq: Fixed!

kvashchuka (approved):
good job!

PR #451, opened 2024-01-31
Fix komodo-snyk-test reporting odd package versions

jonathan-eq:
This PR fixes the issue where the script reports vulnerabilities in versions not used in the release. An example of this was numpy 1.21.3 vulnerabilities being reported, even though we do not use numpy 1.21.3. The reason this was reported was that it is dependency in patsy, but it is never mentioned that it is an indirect dependency. This PR makes it so that the script filters out vulnerable versions that are not present in the release.


berland (approved):
LGTM

PR #452, opened 2024-01-31
Check if version string is a dict in transpiler

larsevj:
Unsure if I should return None or version in the first else case, or throw an error.

### DIFF
--- a/komodo/lint_maturity.py
+++ b/komodo/lint_maturity.py
@@ -116,8 +116,10 @@ def msg_release_exception(release_basename, release_version):
 
 
 def get_release_version(release_basename, tag_exceptions_release):
-    if release_basename not in tag_exceptions_release:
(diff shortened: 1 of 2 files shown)


komodo/release_transpiler.py
```diff
@@ -79,10 +79,13 @@ def _pick_package_versions_for_release(
...
-            version = versions[rhel_ver][py_ver]
-        elif py_ver in versions:
-            version = versions[py_ver]
+        if isinstance(versions, dict):
+            if rhel_ver in versions:
+                version = versions[rhel_ver][py_ver]
+            elif py_ver in versions:
+                version = versions[py_ver]
+            else:
+                version = versions
```
sondreso, line 88: Can this happen :thinking:

sondreso (approved):
LGTM!

PR #453, opened 2024-01-31
Run lint maturity on the transpiled release file

larsevj:
Now that we again are aiming to build matrix builds (multiple rhel and python versions), then we need to lint on the "flat" transpiled release files and not the "raw" release files.


sondreso (approved):
Very nice!

PR #457, opened 2024-02-02
Remove old build scripts

larsevj:
These files do not seem to be in use anymore. No mention of them in repository.yml. Should remove to avoid clutter in komodo.


berland (approved):
👍🏻

PR #458, opened 2024-02-05
Use the builtin venv instead of virtualenv

### LINKED ISSUE #445: Consider using the built in venv instead of virtualenv
Currently we download and use virtualenv for komodo. Now that we no longer support python2 we could use the built in venv instead. This might play nicer with komodoenv which uses venv. So currently we are mixing virtualenv with venv with komodo <-> komodoenv

larsevj:
Resolves #445


berland:
pinging @pinkwah on this

pinkwah (approved):
:tada:

PR #460, opened 2024-02-06
Make komodo py3.12 compatible

jonathan-eq:


jonathan-eq:
So the integration test failed for python3.12 because there is no numpy version with support for both py3.8 and py3.12..

jonathan-eq:
Will squash before merge

.github/workflows/test.yml
```diff
@@ -5,13 +5,15 @@ on: [pull_request]
 jobs:
   build:
     env:
-      RELEASE: release-py3x.yml
-
+      RELEASE_3_X: release-py3x.yml
+      REPOSITORY_3_X: repository-py3x.yml
+      RELEASE_3_12: release-py3_12.yml
+      REPOSITORY_3_12: repository-py3_12.yml
```
berland, line 11: numpy just serves as an example. You can probably pick any other package instead, and then keep it simpler.

jonathan-eq: Yes, I overcomplicated it a little bit

berland (approved):
Nice, remember to clean up commit history.

PR #461, opened 2024-02-07
Improve pylint coverage

jonathan-eq:


komodo/lint.py
```diff
@@ -47,7 +47,7 @@ def lint_version_numbers(package, version, repo):
     maintainer = package_release.get("maintainer", MISSING_MAINTAINER)
 
     try:
-        logging.info(f"Using {package} {version}")
+        logging.info("Using %s %s", package, version)
```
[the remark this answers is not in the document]
jonathan-eq: Okay, I will add a pylint ignore comment.

pyproject.toml
```diff
@@ -77,17 +77,14 @@ komodo-transpiler = "komodo.release_transpiler:main"
...
     "consider-using-sys-exit",
-    "consider-using-with",
     "deprecated-pragma",
     "duplicate-code",
     "file-ignored",
     "fixme",
     "line-too-long",
     "locally-disabled",
     "logging-not-lazy",
-    "logging-fstring-interpolation",
```
berland, line 90: I would keep this here as a global disable (and maybe comment here that this is not to be removed)

komodo/reverse_dep_graph.py
```diff
@@ -158,20 +158,22 @@ def main():
...
+                    ["display"], stdin=dot_proc.stdout
+                )
+                out = io.TextIOWrapper(
+                    dot_proc.stdin,
+                    encoding="utf-8",
+                    line_buffering=True,
+                )
+                run(args.base_pkgs, args.repo, True, pkg, out)
+                out.close()
+                dot_proc.stdin.close()
```
berland, line 176: I would think that this last close is covered by the context manager.

Perhaps there could be a context manager for the `out` object as well.

berland (approved):
Great. Should probably be squashed.

PR #462, opened 2024-02-07
Refactor CLI

### LINKED ISSUE #463: Refactor and add documentation komodo cli
It is difficult to wrap your head around what is happening in the cli if you are new to the tool. 
The functions are doing too much, and they are poorly documented.
Try splitting them into smaller, self explanatory  functions. This would make it easier to read and test.

jonathan-eq:
Resolves #463

### DIFF
--- a/komodo/cli.py
+++ b/komodo/cli.py
@@ -6,12 +6,12 @@ import sys
 import uuid
 import warnings
 from pathlib import Path
-from typing import List, Tuple
+from typing import Any, Callable, Dict, List, Mapping, Optional, Sequence, Tuple, Union
 
 import jinja2
 from ruamel.yaml import YAML
 
-from komodo import switch
+import komodo.switch
 from komodo.build import make
 from komodo.data import Data
 from komodo.fetch import fetch
@@ -23,6 +23,47 @@ from komodo.package_version import (
 from komodo.shebang import fixup_python_shebangs
 from komodo.shell import pushd, shell
 from komodo.yaml_file_types import ReleaseFile, RepositoryFile
+
+
+class KomodoNamespace(argparse.Namespace):
+    """
+    Komodo argument parser namespace
+    """
+
+    pkgs: ReleaseFile
+    repo: RepositoryFile
+    prefix: str
+    release: str
+    tmp: str
+    downloads: str
+    jobs: int
+    download: bool
+    build: bool
+    install: bool
+    dry_run: bool
+    overwrite: bool
+    cmake: str
+    pip: str
+    pyver: Optional[str]
+    workspace: str
+    extra_data_dirs: List[str]
+    postinst: str
+
+
+def profile_time(msg: str) -> Callable[[Callable[..., Any]], Callable[..., Any]]:
+    def decorator(fun: Callable[..., Any]) -> Callable[..., Any]:
+        def wrapper(*args: Any, **kwargs: Any) -> Any:
+            start_time = datetime.datetime.now()
+            res = fun(*args, **kwargs)
+            timings.append(
+                (msg, datetime.datetime.now() - start_time),
+            )
+            _print_timing(timings[-1])
+            return res
+
+        return wrapper
+
+    return decorator
 
 
 def create_enable_scripts(komodo_prefix: str, komodo_release: str) -> None:
@@ -52,7 +93,7 @@ def create_enable_scripts(komodo_prefix: str, komodo_release: str) -> None:
         )
 
 
-def _print_timings(
+def _print_timing(
     timing_element: Tuple[str, datetime.timedelta],
     adjust: bool = False,
 ) -> None:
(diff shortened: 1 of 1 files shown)


jonathan-eq:
This PR does not add any additional unit tests. Should I add a couple for every new function I have split out?

komodo/cli.py
```diff
@@ -25,6 +25,47 @@
...
+    extra_data_dirs: List[str]
+    postinst: str
+
+
+def profile_time(msg: str) -> Callable[[Callable[..., Any]], Callable[..., Any]]:
+    def decorator(fun: Callable[..., Any]) -> Callable[..., Any]:
+        def wrapper(*args: Any, **kwargs: Any) -> Any:
+            start_time = datetime.datetime.now()
+            res = fun(*args, **kwargs)
+            timings.append(
```
xjules, line 58: where is `timings` coming from?

komodo/cli.py
```diff
@@ -62,190 +103,308 @@ def _print_timings(
...
-        shell(f"mv {args.release + str(tmp_prefix)} {args.release}")
-        with contextlib.suppress(OSError):
-            os.removedirs(f"{args.release + str(tmp_prefix.parent)}")
+    shell(f"mv {args.release + str(tmp_prefix)} {args.release}")
+    with contextlib.suppress(OSError):
+        os.removedirs(f"{args.release + str(tmp_prefix.parent)}")
 
-    if args.build and not args.install:
+
+def exit_on_build_only_and_no_install_run(build_mode: bool, install_mode: bool) -> None:
```
xjules, line 196: Does this need to be a stand alone function?

komodo/cli.py
```diff
@@ -62,190 +103,308 @@ def _print_timings(
...
+        release_path (Path):
+
+    Returns:
+        release_root
+    """
+    abs_prefix = release_path.resolve()
+    return abs_prefix / "root"
+
+
+def generate_releasedoc(
```
xjules, line 215: I'd maybe change the name to `general_release_packages`?

komodo/cli.py
```diff
@@ -62,190 +103,308 @@ def _print_timings(
...
-    tmp_prefix = abs_prefix / args.release / "root"
+
+def exit_on_download_only_and_no_build_run(
+    download_mode: bool, build_mode: bool
+) -> None:
+    if download_mode and not build_mode:
+        sys.exit(0)
+
+
+@profile_time("Fetching all packages")
```
xjules, line 134: I like this decorator :) Nice!

komodo/cli.py
```diff
@@ -62,190 +103,308 @@ def _print_timings(
...
-    if args.postinst:
-        start_time = datetime.datetime.now()
-        shell([args.postinst, release_path])
-        timings.append(
-            ("Running post-install scripts", datetime.datetime.now() - start_time),
-        )
-        _print_timings(timings[-1])
 
-    start_time = datetime.datetime.now()
+def find_and_delete_python_bytecode_files(release_root: Path) -> None:
```
xjules, line 319: No time profiling on this one?

xjules (changes requested):
Looks very good @jonathan-eq ! Just had some small comments

komodo/cli.py
jonathan-eq: I found it easier to read, but that might only be me. Do you think I should revert it? 
I was also considering putting this out in a separate function ```if args.download or (not args.build and not args.install):```. Do you think that would be a good idea?

komodo/cli.py
jonathan-eq: Yes, works fine for me

komodo/cli.py
jonathan-eq: I would maybe call that function 
```
def download_mode_or_unspecified_mode(args) -> bool:
    return args.download or (not args.build and not args.install)
```

and change the main function:
```
if download_mode_or_unspecified_mode(args):
    git_hashes = download_packages(
            args.pkgs.content,
            args.repo.content,
            download_destination=args.downloads,
            pip_executable=args.pip,
        )
    exit_on_download_only_and_no_build_run(args.download, args.build)
```

komodo/cli.py
xjules: Sorry, I meant `generate_release_packages` :smile:

komodo/cli.py
xjules: Yes, why not, the more timings we get the better I think.

komodo/cli.py
xjules: Not sure to tbh. Regardless, you need to specify `global timings` since you modify the content of timings.

komodo/cli.py
xjules: > Do you think I should revert it?

Let's keep it then as I thought it makes the readability low, but the function name is quite self explanatory.

> I was also considering putting this out in a separate function if args.download or (not args.build and not args.install):. Do you think that would be a good idea?

Maybe it would, but the those `helping` functions should have the same semantics. Eg. 
`if is_download_mode(args)` -> 
and the same would go for the one above (?):
`if is_build_only(args): exit(0)` 

but not sure

komodo/cli.py
jonathan-eq: Upon prototyping, we found out you dont need the global keyword when you are appending to a list; only when you are overwriting the variable in the global scope.

komodo/cli.py
jonathan-eq: Changed it now. What do you think?

xjules (approved):
Very nice job @jonathan-eq ! :rocket:

jonathan-eq:
This should not break anything. Fingers crossed!

PR #477, opened 2024-02-19
Remove unused python and path related options from build

larsevj:


jonathan-eq:
@larsevj Can this be closed?

PR #478, opened 2024-02-23
Parse correct python version in build/pypaths

### LINKED ISSUE #479: Pypaths function in build.py does not parse correctly python versions for RHEL8
This does not work for python versions on the format `3.8` where it would return `python3` instead of `python3.8`.
```
def pypaths(prefix, version):
    if version is None:
        return ""
    pyver = "python" + ".".join(version.split(".")[:-1])
    return ":".join(
        [
            f"{prefix}/lib/{pyver}",
            f"{prefix}/lib/{pyver}/site-packages",
            f"{prefix}/lib64/{pyver}/site-packages",
        ],
    )
```

Should either remove this or fix it.

larsevj:
Resolves #479

### DIFF
--- a/komodo/build.py
+++ b/komodo/build.py
@@ -2,6 +2,7 @@
 
 import hashlib
 import os
+import re
 import stat
 import sys
 from pathlib import Path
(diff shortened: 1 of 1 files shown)


andreas-el (approved):
We love regex.

Issue #481, opened 2024-03-13
Printed dict difference is wrong due to missing sorting


berland:
`pprint.pformat` "sorts dictionary by keys" by default, but it only the keys on the first level is sorted apparently. This makes that the difference printed by

https://github.com/equinor/komodo/blob/af3569edd9f18cf3568aa5fa0fb56c9cb6272e35/komodo/symlink/sanity_check.py#L97

easily gets wrong, like here:

```
10:25 $ komodo-check-symlinks symlink_configuration/symlink_config.json 
-  'root_links': ['bleeding',
+  'root_links': ['2023.02-py38',
+                 '2021.11-py36',
+                 '2021.06-py36',
+                 '2019.12',
+                 'deprecated-py36',
+                 '2021.10-py36',
+                 '2023.09-py38',
+                 'deprecated-py2',
+                 '2022.10-py38',
+                 '2021.04-py36',
+                 '2022.08-py38',
+                 'stable',
+                 '2022.02-py36',
+                 'testing',
+                 '2021.12-py36',
+                 '2022.06-py38',
+                 '2023.10-py38',
+                 '2022.09-py38',
+                 '2023.04-py38',
+                 '2022.04-py36',
+                 '2022.12-py38',
+                 '2021.07-py36',
+                 '2023.03-py38',
+                 'stable-py36',
+                 '2020.08-py27',
+                 '2021.08-py36',
+                 'unstable',
+                 '2023.08-py38',
+                 '2021.02-py36',
                  'deprecated',
-                 'deprecated-py2',
-                 'deprecated-py36',
-                 'stable',
+                 '2022.01-py36',
+                 '2022.05-py38',
+                 '2021.09-py36',
+                 '2021.03-py36',
-                 'stable-py36',
?                           ^^

+                 'stable-py2',
?                           ^

-                 'testing',
-                 'unstable',
-                 '2019.12',
-                 '2020.08-py27',
?                     ^  ^   ^^

+                 '2023.05-py38',
?                     ^  ^   ^^

+                 '2021.05-py36',
+                 '2023.06-py38',
                  '2021.01-py36',
-                 '2021.02-py36',
?                      ^^^    ^

+                 '2022.11-py38',
?                     ++ ^    ^

-                 '2021.03-py36',
-                 '2021.04-py36',
-                 '2021.05-py36',
?                     ^

+                 '2022.05-py36',
?                     ^

-                 '2021.06-py36',
-                 '2021.07-py36',
[...]

PR #482, opened 2024-03-13
Handle non-sorted lists in dicts when comparing dicts

berland:
This makes the printed diff from the sanity checker make sense

Solves #481

### DIFF
--- a/komodo/symlink/sanity_check.py
+++ b/komodo/symlink/sanity_check.py
@@ -91,10 +91,21 @@ def assert_root_nodes(link_dict):
         )
 
 
-def _compare_dicts(dict_1, dict_2):
+def _sort_lists_in_dicts(_dict: dict) -> dict:
+    result = {}
+    for key, value in _dict.items():
+        if isinstance(value, list):
+            result[key] = sorted(value)
+        else:
+            result[key] = value
+    return result
+
+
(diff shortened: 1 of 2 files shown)


jonathan-eq:
@berland How does the sanity checker output look after this commit?

komodo/symlink/sanity_check.py
```diff
@@ -91,10 +91,21 @@ def assert_root_nodes(link_dict):
         )
 
 
-def _compare_dicts(dict_1, dict_2):
+def _sort_lists_in_nested_dicts(_dict: dict) -> dict:
+    result = {}
+    for key, value in _dict.items():
+        if isinstance(value, list):
+            result[key] = sorted(value)
+        else:
```
jonathan-eq, line 99: Will there ever be a situation where there is another layer of dict? Should we check for this, and use this function recursively?

jonathan-eq (changes requested):

komodo/symlink/sanity_check.py
berland: I will remove `nested` from the function name. It was originally a recursive sorter, but that was before I understood the sorting thing was about lists..

jonathan-eq (approved):
LGTM!

PR #488, opened 2024-04-16
Remove unused komodo transpiler combine mode

### LINKED ISSUE #469: Consider removing komodo transpiler combine
I do not see a use case where we would need the functionality of komodo combine, and it should be removed as to not maintain more code than necessary.

jonathan-eq:
Resolves #469


xjules (approved):
Let's get rid of it. :rocket:

PR #492, opened 2024-04-29
Remove functions from func_def only used in resdata

### LINKED ISSUE #396: Remove conan support functions
These two functions are only used by `ecl`. They have been merged into `ecl` in https://github.com/equinor/ecl/pull/907 and is thus possible to purge from this repository.

https://github.com/equinor/komodo/blob/46ea9b4516f6679af7f8982b43d799d2e0280e28/ci/func_def.sh#L1-L23

Beware that this purging will break komodo builds of ecl releases older than the above PR in ecl.

larsevj:
Resolves #396

### DIFF
--- a/komodo/pypi_dependencies.py
+++ b/komodo/pypi_dependencies.py
@@ -48,7 +48,7 @@ class PypiDependencies:
     ) -> None:
         self.python_version = python_version
         environment["python_full_version"] = python_version
-        environment["python_version"] = python_version[: python_version.rindex(".")]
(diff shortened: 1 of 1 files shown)


pyproject.toml
```diff
@@ -124,6 +124,7 @@ select = [
   "I",   # isort
   "B",   # flake-8-bugbear
   "SIM", # flake-8-simplify
+  "F",   # pyflakes
```
berland, line 127: split in different commit (in case we revert the ecl stuff..)

PR #493, opened 2024-04-29
Replace usage of ntpath with pathlib

### LINKED ISSUE #319: Remove import ntpath
https://github.com/equinor/komodo/blob/72d49dcbdebdcbd0ab709b65dfb424c1c95084ea/komodo/symlink/suggester/release.py#L1

`ntpath` is imported, while the docstring for `ntpath` says:

"Common pathname manipulations, WindowsNT/95 version.

Instead of importing this module directly, import os and refer to this
module as os.path."

If we have an obscure reason to use this, we should document it.

larsevj:
Resolves #319

### DIFF
--- a/komodo/symlink/suggester/release.py
+++ b/komodo/symlink/suggester/release.py
@@ -1,12 +1,6 @@
 import datetime
-import ntpath
-import os
 import re
-
-
-def path_leaf(path):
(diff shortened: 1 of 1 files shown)


andreas-el (approved):
Nice! ✈️

PR #506, opened 2024-07-03
Remove latest package alias feature

larsevj:
- Pip 24.1 removed the legacy feature that the get_latest_pypi_version functionality depended on. We therefore remove this functionality as it is currently only used by 2 packages.
- Also removes the `shell` pypi package from dependencies. This seems to be a wrapper around subprocess, and I cannot find a usage of it ever in the history of komodo. The package also has not been updated in 7 years.


andreas-el (approved):
Proper work! 🌴

PR #524, opened 2024-12-19
Reintroduce py38 support

HakonSohoel:
Reverts the commit that removed python 3.8 support, since it is still needed for the December release


berland (approved):
♻️

PR #525, opened 2025-01-10
Add numpy builds

andreas-el:


komodo/release_transpiler.py
```diff
@@ -2,31 +2,14 @@
...
-from typing import Dict, Sequence, Union
+from typing import Dict, Union
 
 import yaml
 
 from komodo.matrix import format_release, get_matrix
 from komodo.prettier import load_yaml, write_to_file
 
 
-def get_py_coords(release_base: str, release_folder: str) -> Sequence[str]:
```
berland, line 13: The commit message is hiding the fact that you are actually removing functionality (but the functionality seems unused, so it is warranted)

komodo/matrix.py
```diff
@@ -6,35 +6,53 @@
...
+from typing import Dict, Iterator, Optional, Sequence, Tuple
 
 
 def get_matrix(
     rhel_versions: Sequence[str],
     py_versions: Sequence[str],
-) -> Iterator[Tuple[str, str]]:
+    other_components: Optional[Dict[str, Sequence[str]]] = None,
+) -> Iterator[Tuple[str, str, str]]:
     """Return tuples of rhel version and Python version, representing the
```
berland, line 17: update docstring?

komodo/matrix.py
```diff
@@ -6,35 +6,53 @@
...
     current release matrix.
     """
-    for product in itertools.product(rhel_versions, py_versions):
-        rh_ver, py_ver = product
-        yield (f"rhel{rh_ver}", f"py{str(py_ver).replace('.', '')}")
 
+    if other_components:
+        ((component_name, component_seq),) = (
```
berland, line 22: `assert len(other_components) == 1`?

komodo/matrix.py
```diff
@@ -6,35 +6,53 @@
...
 import re
-from typing import Iterator, Sequence, Tuple
+from typing import Dict, Iterator, Optional, Sequence, Tuple
 
 
 def get_matrix(
     rhel_versions: Sequence[str],
     py_versions: Sequence[str],
-) -> Iterator[Tuple[str, str]]:
+    other_components: Optional[Dict[str, Sequence[str]]] = None,
```
berland, line 15: other_components -> other_component unless we plan to support more than three dimensions?

tests/input/test_custom_coordinate_release.yml
```diff
@@ -0,0 +1,26 @@
...
+box:
+  rhel8:
+    py38:
+      numpy1: 1.38.8
+      numpy2: 2.28.8
+    py311:
+      numpy1: 1.311.8
+      numpy2: 2.211.8
+  rhel9:
+    py38:
```
berland, line 20: Is `numpy1` possible at the second level here also, or only at the third level. Since rhel and py works at both levels interchangibly, it is natural to think that this also applies to the third component.

andreas-el: This might be a misconception. I think you can have (in `main`) python-tags, or rhel-tags with python-tags inside.
How we want the end result is obviously up for debate.

komodo/matrix.py
andreas-el: I should limit this to just one extra level. Anything else is ludicrous.

tests/input/test_custom_coordinate_release.yml
berland: Restricting it to only the third level is ok for now!

andreas-el: I've created an issue that describes an possible way of extending the komodo transpiling functionality.
https://github.com/equinor/komodo/issues/526

berland (approved):
go for it! (some squashing only)

PR #536, opened 2025-02-03
Update activator switch to cover other versions and only RHEL8 and upwards

andreas-el:


komodo/switch.py
```diff
@@ -17,24 +18,30 @@
 )
 
 
-def create_activator_switch(data, prefix, release):
+def create_activator_switch(data, prefix: str, release: str):
```
berland, line 21: can we do `data: komodo.data.Data` as well and drop that part in the docstring?

berland (approved):
Looks fine, as in LGTM. Not tested IRL. 🚢 🙏🏻

PR #538, opened 2025-02-04
Introduce the custom coordinate to the activator switch script

andreas-el:


komodo/switch.py
```diff
@@ -7,19 +7,18 @@
 from komodo.data import Data
 
 
-def extract_versions(release_string: str) -> (str, str, str):
+def extract_versions(release_string: str) -> (str, str, str, str):
```
eivindjahren, line 10: May be time to introduce a dataclass for the return type now.

eivindjahren (approved):

komodo/switch.py
eivindjahren: I don't consider this blocking, but could be fixed later.

PR #541, opened 2025-02-05
Update templates to support custom coordinates and valid csh

andreas-el:


tests/test_switch.py
```diff
@@ -28,10 +28,10 @@ def test_write_activator_switches(tmpdir, release, expected_release, custom_coor
 
 if [[ $(uname -r) == *el8* ]] || [[ $(uname -r) == *el9* ]] ; then
```
eivindjahren, line 29: Lets make it identical later.
```bash
rhel_version_number=$(uname -r | grep -oP 'el\\K[0-9]')
if [[ $? ]] ; then
...

eivindjahren (approved):
Small cleanup later, for making the extraction of `rhel_version_number` the same everywhere, but lets merge the bugfix first.

Issue #518, opened 2024-11-15
Make it possible to specify extras in the package status file


eivindjahren:
We would like to specify that `ert` with the extra specifier `everest` is a public package.

In order to do so, we would like to be able to specify that extra in the `package_status`:

```
ert:
  visibility: public
  maturity: stable
  importance: high
  extras: everest
```

PR #545, opened 2025-02-12
Make it possible to specify requirements in check_unused

eivindjahren:
In order to specify extras in check_unused we need to be able to set the extras field of satisfied for _check_unused . https://github.com/equinor/komodo/issues/518


andreas-el (approved):
💯

PR #549, opened 2025-02-17
Avoid tainting source code repo when running tests

berland:
This will make it possible to run test_release_cleanup.py after running test_release_transpiler_argument_types.py


andreas-el (approved):
Nice one! 🌟

PR #551, opened 2025-02-18
Add feature to install komodo-shims at the end

berland:
komodo-shims thus becomes a magic package name that if found, will always be the last pip-package to be installed. This allows such a package to overwrite endpoints that are also installed by other packages.


andreas-el (approved):
Wonderful! 🥧 🍰

PR #554, opened 2025-02-25
Move installation of komodo-shims to the end

berland:
The previous approach would only possibly work if komodo-shims was a pypi package. It will not be. With this commit, it is installed in a separate and new section, and is installed directly from its downloaded source location.


andreas-el (approved):
Let's do this! We can kick a new bleeding build immediately to verify. 👍

PR #556, opened 2025-02-25
Replace pylint usage with ruff

berland:
Not all exceptions that was needed for pylint is needed for ruff,
meaning they are not identical in their linting, but ruff is so much
faster we will let it decide.


andreas-el (approved):
🔧

PR #557, opened 2025-02-25
Remove pointless overwriting of package_name

berland:
The potentially aliased package name is only relevant for pip installs fetching from pypi, but this is no longer performed in this function, rendering the overwriting of the loop variable pointless.

### DIFF
--- a/komodo/build.py
+++ b/komodo/build.py
@@ -248,8 +248,6 @@ def make(
             msg = "pypi_package_name is only valid when building with pip"
             raise ValueError(msg)
 
-        package_name = current.get("pypi_package_name", package_name)
-
         makeopts = current.get("makeopts", "")
         if extra_makeopts:
             makeopts = f"{makeopts} {extra_makeopts}"


komodo/build.py
```diff
@@ -248,8 +248,6 @@ def resolve(input_str):
             msg = "pypi_package_name is only valid when building with pip"
```
andreas-el, line 248: Why not also remove this? This is guarding the statement you are trying to remove, no?

komodo/build.py
```diff
@@ -248,8 +248,6 @@ def resolve(input_str):
...
             raise ValueError(msg)
 
-        package_name = current.get("pypi_package_name", package_name)
```
andreas-el, line 251: Have you tested this? I just worry that this will affect the other build types somehow.

komodo/build.py
andreas-el: E.g.
```
if "pypi_package_name" in current and make != "pip":
    msg = "pypi_package_name is only valid when building with pip"
    raise ValueError(msg)
```

komodo/build.py
berland: Testing in Azure:  
https://github.com/equinor/komodo-releases/actions/runs/13542199588

komodo/build.py
berland: I kept it in case the user would need a hint to how a correct configuration file should look like. But this is probably the wrong place for that error messsage.

berland: No, this is probably the place to put this error.

komodo/build.py
berland: https://github.com/equinor/komodo-releases/actions/runs/13542599931/job/37846905966

